PR #9, opened 2020-11-16
[codebase] Create ABI groups for packages

bmanga:
Currently conan packages built with different compilers cannot be used in projects using a different compiler, even if the generated binary is ABI-compatible.

Add an option to allow conan packages built with different - but ABI-compatible - compilers to be used together more easily.

### DIFF
--- a/design/###-abi-groups.md
+++ b/design/###-abi-groups.md
@@ -1,0 +1,47 @@
+# Proposal: ABI Groups for packages
+
+| **Status**        | **Proposed/Accepted/Deprecated** |
+|:------------------|:---------------------------------------------|
+| **RFC #**         | ####                                         |
+| **Submitted**     | YYYY-MM-DD                                   |
+| **Dependencies**  | RFC #, #                                     |
+
+---
+
+## Summary
+Add an option to allow conan packages built with different - but ABI-compatible -
+compilers to be used together.
+
+
+## Motivation
+Currently conan packages built with different compilers cannot be used in projects
+using a different compiler, even if the generated binary is ABI-compatible.
+While building all dependencies of a project with the same compiler can have some
+clear benefits, not user scenarios warrant such constraint.
+In fact, this fine-grained compiler requirement easily leads to bloat if the same
+packages are used from projects using different compiler versions.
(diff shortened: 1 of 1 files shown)


memsharded:
Hi @bmanga

Thanks very much for your proposal. This reads like a new feature that could be considered for addition in Conan. Not even necessarily for Conan 2.0, maybe a prototype could be done in a proof of concept earlier. As a new feature it might need a few iterations with real world usage until stabilizing it, which is almost impossible to do by design in a proposal from scratch. 

The process of Conan 2.0 aims to be a bit different, to take existing and adopted Conan implemented tools and consider their limitations, their defaults, already existing better solutions, etc., and design and architect around that previous experience and knowledge. As you can see, we have started with very basic things, for example Conan works fine with Python 2.7 as today, but we know that Python 3.6 is a much better alternative in many senses, that will help to maintain the codebase cleaner and more robust, to save lines of code and testing. But this is an architectural decision we don’t want to take without the community support, as if we remove Python 2.7 support in Conan 2.0, introducing it back later will be almost impossible. Adding the feature you propose might be doable in Conan 1.X or why not, later in Conan 2.1, 2.2...

Then, we would like to suggest closing this PR and convert it to a feature request in https://github.com/conan-io/conan. We can label it accordingly, and share similar proposals with the Tribe on a periodic basis so everybody will still be able to give feedback. We will be discussing later in Tribe 2.0 about “package_id”, default ABI compatibility, etc., so there will be opportunities to talk about these things. There will be a nurture process from the Conan repo, from which mature proposals with community support will be brought to the Tribe repo.

bmanga:
Ah alright, sorry about that! 😅

monsdar:
Is there an issue with that in the Conan Issues? We're finally making the leap to VS2019 and it being downwards-compatible to VS2015 is a great help when it comes to thirdparty we're using. Would like to see such a feature implemented.

If there is please link these discussions together, if there isn't I'd be happy to create one.

bmanga:
Here is the one I opened: https://github.com/conan-io/conan/issues/8074
However, I never got around to replying to the comments that were made there.

PR #13, opened 2020-11-25
Migration to Conan 2.0 general guidelines

memsharded:
This proposal summarizes at a very high level some principles that will be applied in the Conan 2.0 design in order to provide an easy upgrade path from Conan 1.X to Conan 2.0. In short:

- Conan recipes will implement a subset of syntax and features in Conan 1.X that is fully operational and can run in Conan 2.0 without requiring changes in conanfiles.
- The command line in Conan 2.0 will change, in a non 1.X backwards compatible way.
- Behavior will change, in a non 1.X backwards compatible way.
- Configuration files might change, in a non 1.X backwards compatible way, but trying to reduce the changes to a minimum.
- When Conan 2.0 is released, it will become the active series, and Conan 1.X will only get important bug fixes.

Please read carefully the contributed proposal document. Please do not discuss specific features that will or not will be breaking, there will be other discussions to talk about specific items in detail.

--------------------------------
Upvote 👍 or downvote 👎 to show acceptance or not to the proposal (other reactions will be ignored)
Please, use 👀 to acknowledge you have read it, but you would be fine with either acceptance or rejection.
Comment and review to suggest changes to all (or part) of the proposal. Specially if you downvote, some comments about the reasons would be appreciated.


design/migration-path.md
```diff
@@ -0,0 +1,92 @@
...
+making the installs a bit less efficient.
+
+There are a few Conan defaults that made sense a few years ago, but not anymore nowadays, like the default ``settings.libcxx=libstdc++``, even
+for gcc>5 compilers. Conan will change this and use ``libstdc++11`` whenever available, but this is breaking behavior.
+
+### Configuration files might change in 2.0
+
+It is very likely that most part of the current configuration files will keep the same, like ``profiles`` or ``settings.yml``.
+But there are other files that might have changes, like ``conan.conf``, which currently contains a lot of variables that don't
+belong there, but to profiles. The way that remotes are defined in the ``remotes.txt`` file might also change.
```
jgsogo, line 69:
```suggestion
belong there, but to profiles. The way that remotes are defined in the ``remotes.json`` file might also change.
```

DoDoENT:
I'm OK with everything in the proposal, but keep in mind that we will need to support using the same machine for working on both conan 1.X and conan 2.X projects, at least until transition period. I don't expect for same recipes to work both in conan 1.x and conan 2.x or even mixing conan 1.x and conan 2.x recipes in the same project, but I think it needs to be possible to work on two different projects - one already using conan 2.x, and another still at conan 1.x.

I am just referring to [my comment](https://github.com/conan-io/conan/issues/1557#issuecomment-696180201) from two months ago 😛

jasal82:
I guess that also means that existing recipes and binary packages might no longer be consumable from a Conan 2.0 build context, is that correct? Will it be possible to tell apart Conan 1.X and 2.X packages in the remotes? Or will there be a new Artifactory repository type for Conan 2.X?

memsharded:
> I'm OK with everything in the proposal, but keep in mind that we will need to support using the same machine for working on both conan 1.X and conan 2.X projects, at least until transition period. I don't expect for same recipes to work both in conan 1.x and conan 2.x or even mixing conan 1.x and conan 2.x recipes in the same project, but I think it needs to be possible to work on two different projects - one already using conan 2.x, and another still at conan 1.x.

@DoDoENT Yes, that will not be a problem. Installing both Conan versions will be possible (virtualenvs most likely, but that can be also discussed), and Conan caches are independent, so it will be possible to have Conan 1.X projects and Conan 2.0 projects at least in different caches (and possibly in the same, that is still unknown).

memsharded:
> I guess that also means that existing recipes and binary packages might no longer be consumable from a Conan 2.0 build context, is that correct? Will it be possible to tell apart Conan 1.X and 2.X packages in the remotes? Or will there be a new Artifactory repository type for Conan 2.X?

Artifactory repo will be the same and compatible with both versions. It might reduce just to have different binaries, that will be identified anyway by different package_ids, so they could live together at the same time. Of course, defining new Artifactory repositories (by users, not a new repository type) is totally doable and might help, but at the moment we don't know if it will be necessary. If something specific is required in the server side for the migration, we will create a new proposal, this one is mainly focused on the client side.

Minimonium:
@memsharded Does it mean that Conan-Center will drop _all_ older recipes at once when Conan 2.0 is released? What about existing local caches, that'd still point to the CCI? Can a mixing of recipes from two different major versions occur during resolving?

memsharded:
> @memsharded Does it mean that Conan-Center will drop all older recipes at once when Conan 2.0 is released? What about existing local caches, that'd still point to the CCI? Can a mixing of recipes from two different major versions occur during resolving?

Nop! It means that the idea is that ConanCenter should transition incrementally and smoothly. We will need to continuously adapt recipes so they are 1.X-2.0 compatible. But the idea is that the recipes are the same. It should be possible to have users using ConanCenter with 1.X and others with 2.0 (the recipes, it might be that not all binaries are available). Until at some point we allow Conan 2.0 new features into ConanCenter recipes, but that will probably be delayed a bit to allow people to upgrade.

uboot:
Just for my understanding: A possible path to migrate an ecosystem of existing recipes to Conan 2.0 would be the following?

1. Limit the syntax/features used in existing recipes to the ones also supported by 2.0 (while still working with Conan 1.x)
2. Switch infrastructure to use 2.0 (ideally no change of recipes necessary)
3. Start to use new syntax/features of Conan 2.x in the recipes

ngerke:
> The command line in Conan 2.0 will change, in a non 1.X backwards compatible way.

Hmm I'm not a big fan of this one. 
That sound like many scripts will break / transition to 2.0 will involve major rework of e.g. ci scripts. Could you elaborate on that please?

jgsogo:
> Just for my understanding: A possible path to migrate an ecosystem of existing recipes to Conan 2.0 would be the following?
> 
> 1. Limit the syntax/features used in existing recipes to the ones also supported by 2.0 (while still working with Conan 1.x)
> 2. Switch infrastructure to use 2.0 (ideally no change of recipes necessary)
> 3. Start to use new syntax/features of Conan 2.x in the recipes

From the POV of ConanCenter it is very important to have a non-breaking syntax to allow a smooth transition that can take months. The same applies to any other _center_ with Conan recipes. 

@uboot , yes, that's the plan for recipes in https://github.com/conan-io/conan-center-index. During the transition, we will use Conan 1.x and Conan 2.0 to generate binaries to check that the recipes are compatible. **After some time (TBD)**, we will stop using 1.x and new features from Conan 2.0 will start to appear, if that happens, new revisions will only work with Conan 2.0.

memsharded:
@ngerke from the proposal document: 

> This is the only possibility. There is a lot of evidence that some commands are not the most intuitive (e.g. the ``search`` should
by default search in remotes, not in the cache. Again, just an example, not something to discuss here), and the only possible way
forward is to change the command line syntax and behavior. Contrary to the Conan recipes, here it is not possible to implement
a "compatible forward" transition, and users will need to upgrade their scripts using Conan to the new Conan 2.0 syntax.

> Some new commands have already started to be designed with some new guidelines for the Conan 2.0 CLI. For example, the ``conan lock`` command uses always the full path to the files, like ``conan lock create conanfile.py``, because the Conan 2.0 will be more explicit, to avoid some confusing ambiguities of the current CLI syntax. But the other commands will need to change, and in most cases, it is impossible to do it in a non-breaking way, for example ``conan install path/conanfile.py`` would be incorrectly processed as a ``pkg_name/version``.


The thing is that if we want to improve the command line and also change some of its legacy defaults, the command line has to be changed, and as we don't have control over the users scripts, what are the alternatives? Another example, right now users doing ``conan install boost/1.70`` to install the package from ConanCenter, get an error telling them that such path doesn't exist. It is necessary to do ``conan install boost/1.70@``, which is not that evident. Also it is necessary to change how information is reported and displayed, too many people relying on parsing the stdout, any possible change there would be breaking too.

markushedvall:
The new CLI is totally understandable and I did vote 👍 But I also think it will be a big problem for us when switching from Conan 1.x to 2.x, and it probably will delay an upgrade. I would have really loved to be able to access the new CLI in 1.X with an environment variable like `CONAN_V2_CLI`.  I totally understand if that is not feasible though.

jgsogo:
@markushedvall,  You nailed it! There is already an environment variable `CONAN_V2_CLI` and Conan will switch to a new command line. Right now there are only a couple of commands implemented (nothing stable, no final arguments names or commands...) but we will keep adding to it and it should help with the transition to Conan v2 😉

markushedvall:
@jgsogo That's great to hear! Thank you for that info 👍

design/migration-path.md
```diff
@@ -0,0 +1,92 @@
...
+define an upgrade path that is as smooth as possible, with a special focus on avoiding the split of the ecosystem like it happened
+with Python 2 and 3.
+
+
+## Proposal
+
+### Conan recipe compatible syntax 1.X => 2.0
+
+This doesn't mean that all Conan 1.X syntax and features will work in Conan 2.0. Many things will be deprecated and removed.
+It means that Conan will provide the features that will be defined as the canonical and recommended ones in Conan 2.0 in 1.X.
```
ytimenkov, line 33:
```suggestion
It means that Conan will provide the features that will be defined as the canonical in Conan 2.0 and recommended ones in 1.X.
```

design/migration-path.md
```diff
@@ -0,0 +1,92 @@
...
+These features will be declared stable in 1.X when there is enough usage evidence, to allow companies to safely adopt them.
```
ytimenkov, line 34: It's a bit unclear what does "Conan 1.X feature" mean: is it Conan 1.35+ (or whatever it will be at that moment) or Conan 1.0?
I quite often find that even Conan 1.x recipes are not compatible and just fixing the recipe using documentation for reference breaks it for some Conan 1.6 (which we've started with). Among examples are tuples vs lists (in settings or options) or using dicts for `default_options`.
And such old versions fail with obscure message like "Can not fetch dependencies" or so.

And I suspect many can still use such ancient versions (at least for some parts) because of build system rigidity or mistakes made during learning.

So my suggestion is to clarify what does "Conan 1.X" mean: Is it a most basic subset which works from 1.0 and onward or is it a specific version which users have to upgrade to, which supports both 1.0 syntax (with already some features deprecated) and subset of 2.0.

Like in python it was only 2.7 which was "source-compatible" with python 3.

ytimenkov (approved):

jason-d-walter:
> > The command line in Conan 2.0 will change, in a non 1.X backwards compatible way.
> 
> Hmm I'm not a big fan of this one.
> That sound like many scripts will break / transition to 2.0 will involve major rework of e.g. ci scripts. Could you elaborate on that please?

Same here.  A command line change would affect scripts setup to automatically build packages based on some CI/CD processes.  It would be nice to transition the command line to the new / preferred syntax.

Daniel-Roberts-Bose:
This proposal sounds fine and is completely expected when a major version is changed, however may I request some documentation on migrating from 1.x to 2.x with a list of what someone would need to at least look at when attempting to migrate from one version to another. This will save a lot of time and gotcha's for everyone. It doesn't have to be highly in depth, but at least a list of things to look at changing, although in depth is always better.

design/migration-path.md
```diff
@@ -0,0 +1,92 @@
...
+- Behavior will change, in a non 1.X backwards compatible way.
+- Configuration files might change, in a non 1.X backwards compatible way, but trying to reduce the changes to a minimum.
+- When Conan 2.0 is released, it will become the active series, and Conan 1.X will only get important bug fixes.
+
+
+## Motivation
+Conan 2.0 is a great opportunity to get rid of some existing legacy, outdated or not optimal defaults, and deprecated practices
+and features. The only way to achieve this is by introducing some breaking behavior. On the other hand, it is necessary to
+define an upgrade path that is as smooth as possible, with a special focus on avoiding the split of the ecosystem like it happened
+with Python 2 and 3.
```
datalogics-kam, line 25: I think given the listing of "non 1.X backwards compatible" changes above, this is going to set up exactly the split of the ecosystem that happened with Python 2 and 3.

For small projects it looks easy, but for enormous projects in enterprise applications, it'll delay 2.0 adoption. When you have to adapt tools that are wrapped around Conan, change CI, or have to plan to change everything at once, it just becomes painful enough to not migrate.

I think it's much better to evolve rather than break. Keep old commands working, but supply compatible fixups. Make sure the underlying structures (repositories, cache, recipes) are compatible.

datalogics-kam:
This is a potentially severely disruptive change. We've got a lot of large projects, some of which only get attention quarterly or less, and this would put us in the state of having legacy projects on Conan 1.x until we can allocate time to get them to 2.x.

I think we really need to look at the various levels of compatibility and figure out how to best resolve them, or figure out ways of being compatible across versions. I think this means probably separate proposals.

- Command line compatibility.
- Conan API compatibility (yes, we use that in our wrapper tools)
- Local cache compatibility
- Remote repository (Artifactory) compatibility
- Recipe compatibility
- Binary package compatibility
- Availability of 1.x/2.x compatible packages from Conan Center Index.

Python 3 came about on the theory of making breaking changes, and getting it all over once. With hindsight, I don't think they'll do that again.

I think evolving the interface is preferable to choosing a breaking point and then breaking a lot of things. So I vote this one down.

design/migration-path.md
memsharded: > I think it's much better to evolve rather than break. Keep old commands working, but supply compatible fixups. Make sure the underlying structures (repositories, cache, recipes) are compatible.

Of course it is, but we have reached one point that we are struggling to evolve. For example, the amount of complicated logic and code to deal both with revisions and without revisions, is so huge, that evolving things take a lot of time, and what is more important, it is also very risky, and we inevitably break too many things for even small improvements and changes. So this is the reason we are proposing to remove support for not using revisions (see the other proposal), and this will be inevitably breaking.

rjktcby:
is it maybe possible to guarantee that packages built and uploaded with Conan 2.0 will still be consumable with Conan 1.x? In that case I can imagine gradually upgrading CI scripts starting from inner layer libs and going up to libs/applications that depend on them later.

memsharded:
@rjktcby 
> is it maybe possible to guarantee that packages built and uploaded with Conan 2.0 will still be consumable (as a requirement, or bulid_requirement) with Conan 1.x? 

Not necessarily, because for example, Conan 2.0 might propose a new package_id computation. But as long as the recipe is compatible (and that is a goal), it will be fine, because the idea is that you can build that package with both 1.0 and 2.0 and each Conan version will have its own binaries. So even if the binary is not exactly the same one, as long as you can use it from both Conan versions, you will be able to migrate smoothly (modulo the breaking changes that might happen in the command line syntax)

design/migration-path.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Behavior will change in 2.0
+
+The fact that recipes will be forward compatible doesn't imply that the behavior will be the same. Some things might change,
+for example (again, as an example, not to be discussed here), the default ``package_id_mode`` might change, and then it might become necessary
+building new binaries. Or the ``build_requires`` recipes will always be fetched, even if they are not needed for building from source,
+making the installs a bit less efficient.
+
+There are a few Conan defaults that made sense a few years ago, but not anymore nowadays, like the default ``settings.libcxx=libstdc++``, even
+for gcc>5 compilers. Conan will change this and use ``libstdc++11`` whenever available, but this is breaking behavior.
```
dean0x7d, line 63: I'm in favor of the overall document and I can see most of the transition going pretty smoothly (with the measures outlined in the document and the PR comments). The only thing that I'd really want to watch out for is this section: behavior changes for identical recipes. This is because it has the potential for silent breaks. If the recipe syntax changes, it's going to explicitly fail and that's obvious. However, if the same recipe does something different after the version change, that's harder to detect.

I can see these kinds of changes being necessary but it's definitely the main aspect I'll be watching out for. The things that would be good from my point of view:
* Document these clearly, including a transition guide.
* When possible, ensure that the new behavior is stricter than the old one. For example, going to a stricter default `package_id_mode` may cause some inefficiencies during the build process but it will not cause any ABI issues. That's fine. The other way around, if the new behavior became less strict, would result in silent ABI problems for anyone who missed that point in the transition docs. (My example here may be a bit contrived. Based on the doc, I assume you're already going in a stricter direction.)

jcar87:
Oh I have a few comments:

- I know conan currently supports the `CONAN_V2_MODE` setting, but my understanding is that it errors out or forbids behaviour where conan v2.0 has different defaults. I  wonder if one could have (future) 1.x releases issue deprecation warnings for things that are known to be going away on v2.0, without necessarily issuing any errors. As a user I would like to enable a "warn me about anything that will be deprecated in 2.0". I could maybe do a "full run" on CI, and capture these warnings, and get a "global view" of where our migration efforts need to be, rather than the potentially tedious "fix this error, try again, fix this other error, try again". 

- I do echo the concerns of having behaviour changes where conan 2.0 does something slightly differently than 1.0 for the same still-compatible code (e.g. in recipes) - silent behaviour changes could be tricky to handle.

- related to the concern of existing recipes (1.x) being compatible with 2.0, I presume there's a case where at some point conan 2.x will introduce features in the recipes that conan 1.x is no longer compatible with - I can see from comments above that there _is_ an interest to keep the same recipes compatible with clients 1.x and 2x for a while, but what happens after that? In particular i'm interested in how would a 1.x client would detect the situation where it should no longer even attempt to parse the recipe?

- and the other way around, let's say 2 years from now I'm happily using conan 2.x and all the recipes in the current head of my development branch are all 2.x compatible - what if for some reason (with nonzero probability!) I have to try and build an older commit where v1.x recipes where a thing? Would that future client be able to to parse and correctly work with legacy 1.x recipes? If the answer is no and I have to install conan 1.x on a separate virtualenv, would that corrupt my conan settings / database / local cache in any way?

ytimenkov:
What others wrote about ability to use binary packages from different versions made me thinking:

Where CI involved it is usually 1 pipeline building (and testing) a certain recipe. So in terms of upgrade most likely pipelines will be upgraded one by one (at least not all at ones). This means that having compatible binaries is much more important than recipes in such scenarios: once a certain component switched to conan v2 it crossed the border and its packages can't be used by v1 (and actually vice versa).

Not being able to simply convert binary packages means duplication of CI for a very unpredictable time (who knows when some backport will be needed...).

memsharded:
> I know conan currently supports the CONAN_V2_MODE setting, but my understanding is that it errors out or forbids behaviour where conan v2.0 has different defaults. I wonder if one could have (future) 1.x releases issue deprecation warnings for things that are known to be going away on v2.0, without necessarily issuing any errors. As a user I would like to enable a "warn me about anything that will be deprecated in 2.0". I could maybe do a "full run" on CI,

The current behavior we are implementing is exactly that: If CONAN_V2_MODE is not enable => Warn. If it is enabled => raise. I think this is what you want, try to enable it today!

> I do echo the concerns of having behaviour changes where conan 2.0 does something slightly differently than 1.0 for the same still-compatible code (e.g. in recipes) - silent behaviour changes could be tricky to handle.

Totally understandable. Still, there might be some cases that this might be unavoidable. A simple example could be if we change the default ``package_id_mode``. Exactly the same ``package_id()`` method in one recipe, with the same ``conan.conf`` configuration, will still result in a different ``package_id``, with a possible different behavior (needing to build it from source for example). In the same way, the current approach to handle symlinks in packaging is problematic. We might want to change the behavior of ``self.copy()``, and move part of its implementation to a helper ``tool``. The ``self.copy()`` call will be syntax-compatible, but if you don't provide the previous call to the corresponding ``tool`` that handles the symlinks, before calling ``self.copy``, then you will still get a different behavior. We will try to minimize these, and document everything that changes, but it seems that we cannot guarantee behavior compatibility if we want to improve a few things.

> related to the concern of existing recipes (1.x) being compatible with 2.0, I presume there's a case where at some point conan 2.x will introduce features in the recipes that conan 1.x is no longer compatible with - I can see from comments above that there is an interest to keep the same recipes compatible with clients 1.x and 2x for a while, but what happens after that? In particular i'm interested in how would a 1.x client would detect the situation where it should no longer even attempt to parse the recipe?

We already have the mechanism ``required_conan_version``. When recipes start to use Conan 2.X features only, they should update their ``required_conan_version`` information, so Conan 1.X clients can raise safely and informing of the reason. Conan 1.X have been supporting it for a while, so they will be able to raise safely

> and the other way around, let's say 2 years from now I'm happily using conan 2.x and all the recipes in the current head of my development branch are all 2.x compatible - what if for some reason (with nonzero probability!) I have to try and build an older commit where v1.x recipes where a thing? Would that future client be able to to parse and correctly work with legacy 1.x recipes? If the answer is no and I have to install conan 1.x on a separate virtualenv, would that corrupt my conan settings / database / local cache in any way?

If you were using the subset of syntax of Conan 1.X that was compatible with Conan 2.0, you might be able to build it with the new Conan 2.X client. But as the behavior, configuration, etc. might still be different (check the other proposal=using revisions always, your 1.X build was probably not using revisions), then you really need a Conan 1.X client. And with it you will get its environment, configuration, cache (you will need to define its CONAN_USER_HOME to not conflict with other Conan 2.0 features), etc. I think it is the only possible way to go back in time to reproduce a Conan 1.X build.

PR #14, opened 2020-11-25
[general] Make revisions enabled the default for Conan 2.0

czoido:
Package revisions have proved to work very well for achieving packages traceability and immutability and also maintaining the two different modes is challenging. As explained in the proposal, we think we should enable package revisions as the default and only behavior, and remove support for packages without revisions in Conan 2.0.

---

### DIFF
--- a/design/###-general-revisions.md
+++ b/design/###-general-revisions.md
@@ -34,3 +34,7 @@ Use revisions enabled as the default mode in Conan 2.0 and remove support for pa
 revisions:
 - All commands will assume that packages are using revisions, and behave accordingly.
 - Produced files, like lockfiles, will contain the package revisions automatically.
+
+## Detailed Design
+Removing support for packages without revisions, means that Conan servers should use revisions as
+well (Rest HTTP V2 support).
--- a/design/###-general-revisions.md
+++ b/design/###-general-revisions.md
@@ -1,0 +1,36 @@
+
+# Proposal: General - Make revisions enabled the default for Conan 2.0
+
+| **Status**        | **Accepted**                                      |
+|:------------------|:--------------------------------------------------|
+| **RFC #**         | ###                                               |
+| **Submitted**     | 2020-11-25                                        |
+| **Tribe votes**   |                                                   |
+
+---
+
+## Summary
+Enable [package revisions](https://docs.conan.io/en/latest/versioning/revisions.html) as the default
+and only behavior, and remove support for packages without revisions in Conan 2.0.
+
+## Motivation
+Package revisions have proved to work very well for achieving packages traceability and immutability.
+Also, combined with [lockfiles](https://docs.conan.io/en/latest/versioning/lockfiles.html), this
+feature is a powerful tool for achieving reproducible builds.
+
+Maintaining two different modes in the current version is challenging for several reasons, like:
+- Exactly the same command  may have different meaning and behavior depending on if revisions are
+  enabled or not. That is the case of the `--update` argument for the `conan install` command. This
+  fact can sometimes confuse users about how Conan should behave in certain situations.
+- Each mode has its version of the server API, v1 for packages without revisions and v2 for revisions
+  enabled. Besides the fact that the v2 API supports more features, making the server API v2 the
+  default would make the maintenance simpler.
+- Full decentralization of packages on the server side is impossible without revisions as a result of
+  packages not being immutable, the same package reference can be completely different thing in
+  different servers.
+
+## Proposal
+Use revisions enabled as the default mode in Conan 2.0 and remove support for packages without
+revisions:
+- All commands will assume that packages are using revisions, and behave accordingly.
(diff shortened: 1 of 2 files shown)


steinerthomas:
So maybe the `default_package_id_mode` will also change in the future? But I guess this will then be discussed in another pull request?

markushedvall:
We have already activated revisions. And when we have recipe and source code in the same repository we use `revision_mode:  "scm"`, which is very helpful to connect a revision to a git commit.

memsharded:
@steinerthomas yes, the ``default_package_id_mode`` might change, but still not decided, it will be a future proposal. This proposal doesn't affect the package_id, using revisions do not change the package_id, unless you do it in conf or ``package_id()`` method.

foundry-markf:
We too have activated revisions this year, and are appreciating being able to pin to specific revisions (both recipe and package). We're also migrating recipes from Perforce to git, to take advantage of the scm revision_mode out of the box support, to tie recipe revisions in Artifactory directly to the git SHA.

With recipes that still remain in Perforce, we've had to fall back to the hashing mode for recipe revisions. This has been seen as problematic in edge cases, such as when rolling back recipe changes to earlier recipe contents, as hash collisions occur and it's not possible (thirdparty information here) to re-upload an existing recipe revision on a more recent date. (Resolved by having to go into Artifactory and removing unwanted revisions).

Don't have that problem in git, as we don't allow changing history.

mathbunnyru:
Could you please explain, what does this change break for the existing users?

memsharded:
> Could you please explain, what does this change break for the existing users?

@mathbunnyru, this shouldn't be a massive break, but it might break some users that are relying on some of the (bad) behaviors of not using revisions. If you don't use revisions, you can do changes to a recipe or source code, not bump the version, and create a package again. It will overwrite the existing package. It you upload it to the server, it will overwrite the server package. If you are creating tons of packages, you might need to implement some cleaning policy in the server, otherwise the storage needed might grow too much. It might break users that for example use the known paths in Artifactory server to download things directly without using Conan. It might also affect users that use some server that does not implement the api v2 yet.

mathbunnyru:
Thanks @memsharded.
Sound good for me.

ytimenkov:
Unfortunately I haven't used package revisions in practice to see the whole potential.
And here I mean the whole pipeline with `default_package_id_mode` and rebuilding packages when dependencies' ids change and proper lockfile handling.
(`scm` attribute and immutability are widely used even with v1 API).

To make things more complicated not all servers support v2 API (GitLab for example...) which makes migration bit more complicated.

However I do believe in the potential of revisions (and probably infrastructure issues will be resolved by the time Conan 2.0 is out) so I don't downvote the proposal 🙂.

puetzk:
I have no problem with "default", but one small heartburn with "only"

> A client with revisions enabled will only find binary packages that belong to the installed recipe revision

It has, on occasion, been necessary to fix something in the conanfile.py that was broken by a newer version of conan and/or a newer python. It's rare, and usually it implies something a mistake was gotten away with in the first place, but it has happened.
Even today, this makes the binaries "outdated", but that has little consequence, and can be fixed by updating the conaninfo.txt [recipe_hash] and conanmanifest.txt, then doing such a "nuke" upload.

It is generally pretty undesirable to actually recompile the binaries of an older release, since that could have rippling downstream effects (especially given the issues discussed elsewhere with package_id and MSVC minor versions, but just on general principles too).

Might it be possible to have some way something (maybe as part of `conan copy`?) to explicitly bring existing binaries over to a newer recipe revision? This is a lot like like using export-pkg as in https://docs.conan.io/en/latest/creating_packages/existing_binaries.html#packaging-pre-built-binaries, except without even using the package() step of the recipe, since the files are already copied into the destination layout.

I think that would make this a straightforward win - you'd never accidentally get old binaries, the server part would be immutable, lockfiles would work as intended -- and one could still deliberately, but straightforwardly, carry binaries forward when there is a a recipe fix that only affects the consumer side, not the producer.

memsharded:
> Might it be possible to have some way something (maybe as part of conan copy?) to explicitly bring existing binaries over to a newer recipe revision? This is a lot like like using export-pkg as in https://docs.conan.io/en/latest/creating_packages/existing_binaries.html#packaging-pre-built-binaries, except without even using the package() step of the recipe, since the files are already copied into the destination layout.

Hi @puetzk, I think this could be a valid feature, and I like that it might not break the immutability. There could be different strategies to implement this (conan copy not one of them, because that copies from the recipe, but would need to be something different). It could even be very efficient if implemented server side. It is necessary though to fix a few things first, for example removing the recipe hash from the conaninfo.txt in the packages. But that might be possible in Conan 2.0. If you want to submit this as a feature request (to the Conan repo), I will label it as Conan 2.0 and take it into consideration.

czoido:
Thank everyone for the participation!
As the votes reveal a clear consensus on making revisions the default mode for Conan 2.0 we are going to merge the proposal's PR.

PR #17, opened 2020-12-18
proposal: lowercase references

memsharded:
Recipes references (package name, version, user and channel) should be lowercase. They will continue accepting other characters, digits(0-9), underscore, dots, etc. but letters (a-zA-Z) should be always lowercase. Trying to create a package with uppercase will raise an error (A temporary opt-out will be enabled to allow teams to migrate their packages gradually)

This proposal tries to remove some of the problems with package different casing like Boost, boost, BOOST:

- Lack of conflicts and impossibility to do overrides
- OSs specific issues as Windows not able to handle different casing packages
- Security issues and typo-squatting attacks.

---

### DIFF
--- a/design/017-lowercase-references.md
+++ b/design/017-lowercase-references.md
@@ -1,10 +1,10 @@
 # Proposal: References will be lowercase
 
-| **Status**        |                            |
-|:------------------|:---------------------------|
-| **RFC #**         |                            |
-| **Submitted**     | 2020-12-18                 |
-| **Tribe votes**   |                            |
+| **Status**        | **Accepted**                                      |
+|:------------------|:--------------------------------------------------|
+| **RFC #**         | [017](https://github.com/conan-io/tribe/pull/17)  |
+| **Submitted**     | 2020-12-18                                        |
+| **Tribe votes**   | :thumbsup: (47) :thumbsdown: (3) :eyes: (0)       |
 
 ---
 
--- a/design/lowercase-references.md
+++ b/design/lowercase-references.md
@@ -45,7 +45,10 @@ All package references, (including the name, version, user and channel) should b
 
 Then, Conan 2.0 will enforce recipes references (package name, version, user and channel) to be lowercase. They will continue accepting other characters, digits, underscore, dots, etc. but letters inside references should be always lowercase. Any attempt to create a package that is not lowercase will fail. If other mechanisms are employed (directly editing the Conan cache, or changing the server names) to change the reference of any package, that will be a broken one with undefined behavior.
 
+## Detailed Design
+
+One of the challenges of using packages with different casing for same name is the conflicts they generate while installing in the Conan cache in some operating systems. Conan 2.0 will provide a new cache layout that will be able to store these packages with different casing without conflicts. They will still be different, independent packages, and it will not be possible to use them in the same graph, as they will not conflict in the graph, but produce duplicated symbols while linking, but it will be possible to be installed in the cache, highly facilitating the migration to lowercase packages.
 
 ## Migration notes
 
-As changing package names might take some time, Conan 2.0 will introduce a temporary opt-in in conan.conf to allow creating packages non-lowercase. This temporary opt-in doesn’t imply any behavior, all the above limitations of Conan 1.X will persist, and Conan will stop doing checks for OS specific issues, like the Windows one. This opt-in will be removed in later Conan 2.X version.
+As changing package names might take some time, Conan 2.0 will introduce a temporary opt-in in conan.conf to allow creating packages non-lowercase. This temporary opt-in doesn’t imply any behavior, Conan will still be case-sensitive, so they will be treated as completely different packages. But as commented above, the new cache will allow to install them without conflict. This change will make the migration to lowercase package names exactly the same as a major version bump, which is a well known process that can be done with reasonably planning. This opt-in will be removed in later Conan 2.X version, at least 6 months after the Conan 2.0 release (which will still require at least 6 months).
--- a/design/lowercase-references.md
+++ b/design/lowercase-references.md
@@ -1,0 +1,51 @@
+# Proposal: References will be lowercase
+
+| **Status**        |                            |
+|:------------------|:---------------------------|
+| **RFC #**         |                            |
+| **Submitted**     | 2020-12-18                 |
+| **Tribe votes**   |                            |
+
+---
+
+
+## Summary
+
+Recipes references (package name, version, user and channel) should be lowercase. They will continue accepting other characters, digits(0-9), underscore, dots, etc. but letters (a-zA-Z) should be always lowercase. Trying to create a package with uppercase will raise an error (A temporary opt-out will be enabled to allow teams to migrate their packages gradually)
+
+
+## Motivation
+
+Conan 1.X is case sensitive, Boost, boost, BOOST are all different package names corresponding to completely different things. They cannot be overridden, they will not conflict with each other (in the graph sense, they will certainly produce errors at link time if they provide the same library names and symbols).
+
+The server side will store them as different items, they will be retrieved as different packages with “conan search” command, and they can be installed independently with “conan install”.
+
+This behavior generates a lot of issues, bugs and problems:
+
+### Lack of conflicts
+
+If package names are case sensitive, then it is possible that one package depends on a given boost/1.72 package and another one depends on BOOST/1.72 package. If they are different packages, the graph will be correct, and the errors will appear at build time, when library and symbol conflicts will happen.
+
+As they are different, it is not possible to do overrides, and trying to force a common package from the downstream consumers will fail.
+
+If the casing is not enforced, it is a matter of time that different communities generate packages for the same libraries (Boost, Poco, OpenCV, etc), that just differs in the casing, making them unusable (or at least very hard to swap), for users that were requiring packages with other casing.
+
+### OS specific issues
+
+Some OS, like Windows, have case-insensitive filesystems, making it impossible in practice to store different packages with different casing in the Conan cache.
+
+### Security issues
+
+Typo-squatting is a very well known attack vector for package managers. Having a package that just slightly differs from the “official” one and resolves differently is a way to deploy in developer machines different kinds of malware. Uppercase/lowercase differences are the simplest and most common differences, because they are not even typos, but a completely valid spelling. Let's say that ConanCenter contains a package called “boost/1.72”, malicious users could upload to other remotes also used by the community (but not as reviewed as ConanCenter) a “Boost/1.72” package just waiting there for users that spell it that way in their recipes.
+
+
+## Proposal
+
+All package references, (including the name, version, user and channel) should be unique, unambiguous and not confusing for users, reducing the cognitive effort. While making the package names case insensitive might alleviate issue number 1 (“Lack of conflicts”), the security issues still persist, and the cognitive overhead for users will still be there.
+
+Then, Conan 2.0 will enforce recipes references (package name, version, user and channel) to be lowercase. They will continue accepting other characters, digits, underscore, dots, etc. but letters inside references should be always lowercase. Any attempt to create a package that is not lowercase will fail. If other mechanisms are employed (directly editing the Conan cache, or changing the server names) to change the reference of any package, that will be a broken one with undefined behavior.
+
+
+## Migration notes
+
+As changing package names might take some time, Conan 2.0 will introduce a temporary opt-in in conan.conf to allow creating packages non-lowercase. This temporary opt-in doesn’t imply any behavior, all the above limitations of Conan 1.X will persist, and Conan will stop doing checks for OS specific issues, like the Windows one. This opt-in will be removed in later Conan 2.X version.


DoDoENT:
I completely agree, but some time for migration is required and during this time we will have to find a solution to have different cases in conan cache.

For example, we have custom package `ZLib/<version>@company/stable` and we would very much like to rename it to `zlib/...`, but this would require conan to support having both `ZLib` and `zlib` in its cache during the transition period, i.e. until all packages are migrated to use `zlib`. How can this be achieved? Currently, conan throws an error if different projects attempt to install different-case names of the same package.

Multiple conan caches? This would definitely confuse most of our junior and less experienced developers, but I guess we could live with it. The Jenkins would also have to be configured to use different cache folders for conan1 and conan2.

It would be ideal if that were managed automatically by conan.

memsharded:
> Currently, conan throws an error if different projects attempt to install different-case names of the same package.

That is the thing, Conan raises in some OS, just to prevent even weirder error messages because that OS (Windows) doesn't support that. As the OS is not supporting that, that would be very challenging, not impossible, but challenging. And specially with a very high risk of breaking something, these kind of changes is typically very risky.

I am not fully sure either that allowing the cache to store both Zlib and zlib will solve the most relevant challenges. Lets say that you have Zlib that is required by LibA, LibB and LibC packages, which are required by App. Then someone creates ``zlib`` lowercase package and uploads it, so far so good. Now if only 1 LibA changes its requires to ``zlib``, even if the cache supports having both ``Zlib`` and ``zlib``, the project will fail anyway. It is necessary to upgrade all the consumers at once, and then only ``zlib`` will be pulled into caches.

If ``Zlib`` is used by different projects, those projects build should typically not share the cache in CI anyway, so that shouldn't be an issue. Then I guess it will be mostly an inconvenience for (Windows) developers, and the solution would be indeed to define 2 Conan caches, one for the project that already migrated to ``zlib`` and the other to the project that still uses the old ``Zlib``.

As we are talking about a long term thing: if this gets approved, starting from now, until Conan 2.0 is released, and a few releases until some Conan 2.X removes the opt-out, we are probably talking about 1 year, I hope it will be time enough to migrate.

DoDoENT:
> I am not fully sure either that allowing the cache to store both Zlib and zlib will solve the most relevant challenges. Lets say that you have Zlib that is required by LibA, LibB and LibC packages, which are required by App. Then someone creates zlib lowercase package and uploads it, so far so good. Now if only 1 LibA changes its requires to zlib, even if the cache supports having both Zlib and zlib, the project will fail anyway. It is necessary to upgrade all the consumers at once, and then only zlib will be pulled into caches.

Not necessarily.

Let's say that `LibA/1.0.0` depends on `ZLib`, `LibB/1.0.0` depends on `LibA/1.0.0`, `LibC/1.0.0` depends on `LibB/1.0.0` and App depends on `LibC/1.0.0`. We want to make the transition to lower-case package names but cannot do all at once.

So, we first create `zlib` and update `LibA/1.0.0` into `liba/2.0.0`. The App still uses `ZLib` and `LibA/1.0.0`, but conan cache (especially on CI) may already contain `zlib` and `liba/2.0.0`.

It will pass some time (weeks, months) until we have `libb/2.0.0` that depends on `liba/2.0.0` and `libc/2.0.0` that depends on `libb/2.0.0` and finally updated App that uses all lower-case packages. In the meantime, we need to find a way to have both `LibA/1.0.0` and `liba/2.0.0` in the same conan cache.

DoDoENT:
Of course, if someone overrides App's dependencies to `liba/2.0.0` before `libb/2.0.0` and `libc/2.0.0` are available, the project build will fail, as you suggested. But that must not prevent gradual transition to lower-case names.

DoDoENT:
Maybe conan 2.0 could solve that problem in the following way:

- keep all packages in `~/.conan/data` as current and forbid any packages with uppercase letters to be put into `~/.conan/data` cache
- if opted-in for compatibility with case-sensitive packages, put all those into `~/.conan/case-sensitive-data` and raise an error if finding a different-case conflict (e.g. Boost vs BOOST) - this is also the current behaviour for Boost vs boost.
- use both `~/.conan/data` and `~/.conan/case-sensitive-data` as package caches during the transition period (i.e. while conan 2.0 is opted-into allowing case-sensitive packages). I guess this is similar to how `short_paths` feature on Windows work. Also, treat different-case packages as different packages: this would allow having both boost and Boost in the dependency graph, but it's up to the user to solve that issue (i.e. by using the `provides` mechanism of conan packages or by avoiding both versions in the dependency graph as I explained in earlier comments).
- after the first install of conan 2.0, migrate all packages with uppercase letters into `~/.conan/case-sensitive-data` and issue a warning that case-sensitive package names are deprecated and will be removed in a future update.

Would that work, @memsharded ?

datalogics-kam:
Another solution to case-insensitive filesystems would be to escape the upper case letters, for instance `Expat` would be stored in `%expat` on disk. Or something.

Side question: Can package names contain arbitrary Unicode characters? We may want to consider either accommodating or prohibiting that, taking a stand one way.

DoDoENT:
> Another solution to case-insensitive filesystems would be to escape the upper case letters, for instance Expat would be stored in %expat on disk. Or something.

This idea was already discussed [here](https://github.com/conan-io/conan/issues/1557). Apparently, such thing would break a lot of things, to [name a few](https://github.com/conan-io/conan/issues/1557#issuecomment-696034015):

> - Build systems will fail when finding unexpected characters like % in the path.
> - Templates, user file manipulation or scripts will fail when finding % in the path.
> - Packages in windows that are named "THIS_PACKAGE_NAME_IS_VERY_LONG", will largely increase its path length, with increased likelihood of hitting the Windows path length limit and failing.
> - User scripts that access the cache will fail, and will re-fail if for some reason we need to change the encoding of the capital letters
> - Conan version downgrades will crash badly.

memsharded:
> Side question: Can package names contain arbitrary Unicode characters? We may want to consider either accommodating or prohibiting that, taking a stand one way.

Package names right now in Conan 1.X are already constrained to (a-zA-Z0-9._-+) IIRC. We plan to maintain that, just removing the A-Z. Encoding is always a nightmare and so far it has not been a strong need, I think we should continue with this constraint.

memsharded:
> after the first install of conan 2.0, migrate all packages with uppercase letters into ~/.conan/case-sensitive-data and issue a warning that case-sensitive package names are deprecated and will be removed in a future update.

@DoDoENT actually, it is very likely that we need a whole redesign of the cache to account for the revisions-always-on new mode, so we certainly will try to accommodate this issue if possible (and not very error prone or fragile, could be worse the cure than the disease...)

steinerthomas:
We will start refactoring our packages to lowercase soon, so no problem!

mathbunnyru:
It might be nice to see the current number of recipes that will not be considered legal after this change - this will give an understanding of how much changes users need to do.

foundry-markf:
We've encountered the casing issue a number of times, both on case insensitive file systems, and also in the ambiguity recipe developers face on 'how do I name this package'? We often try to name the same as what CMake would expect in a `find_package` call (since that is case sensitive), for consistency and least surprise, but happy to move away from that for the extra benefits and simplification.

So, having an enforcement is a good thing, as I see it.

I do worry about the migration though, as has already been discussed here. Having a solution within a local cache redesign itself, as mentioned, sounds interesting.
The alternative, as I see it, is not to do an incremental update as in the example above, but do a big bang update, which would be painful, but feasible. (For us anyway, as we only have private repositories of Conan packages.)

DoDoENT:
> It might be nice to see the current number of recipes that will not be considered legal after this change - this will give an understanding of how much changes users need to do.

139+ packages on our side, with complex inter-dependencies. We definitely cannot afford a big bang update, even though all our packages are hosted on our private Artifactory server.

mathbunnyru:
> 139+ packages on our side, with complex inter-dependencies. We definitely cannot afford a big bang update, even though all our packages are hosted on our private Artifactory server.

I see, that's quite a lot 👍 
Thanks for the answer!

Daniel-Roberts-Bose:
This change would mean our team could not transition to using conan 2 in any short amount of time (if ever). We have probably 250+ internal packages that use camel casing and so this change would break every package, reference, and line of code that uses those package names. 

It would be a monumental effort to find all the places that require this change and it would likely just prevent us from ever moving to conan 2. 

Is there any way to instead implement something that simply ignores casing and always resolves to correct reference?

Croydon:
Maybe a global Conan client configuration, which takes all references and interprets them lower-case. Conan could then print a warning that references are overwritten (e.g. `Warning: ZLib/1.2.11@company/stable was overwritten to zlib/1.2.11@company/stable because of ...` . 

If the lower-case dependencies would not exist you would get the regular "dependency not found" error.

Could be opt-in in a transition phase and then be implicitly true afterwards.

Maybe that would allow to work on making the packages available lower-case step by step, and optionally test if it would work with all-lowercase already.

memsharded:
Hi @Daniel-Roberts-Bose 

> Is there any way to instead implement something that simply ignores casing and always resolves to correct reference?

We have evaluated this, and basically not. It would be a humongous effort, that involves changing servers as well (this means that Conan 2.0 will take at least twice the time), will be very buggy and source of issues forever, and the most important one: in fact it doesn't solve the main issues, being case-insensitive is super problematic as well, it is not that we want to be case-insensitive, and now a ``openssl/xxxx`` requires can resolve to another server ``Openssl/xxxx``, that would be really bad. 

It is important to know that in practice you will have more than 1 year to do this (Conan 2.0, plus the opt-out that Conan 2.0 will implement), you can start today, start changing packages lowly coupled with others, etc. Note that the servers are still case sensitive, so lets say that you have LibA/1.0, and you are creating a new 1.1 or 2.0 version. The problem is no different. The new version, you create ``liba/2.0``, instead of ``LibA/2.0``, and upload to the servers. Nothing will break there, as old consumers will still use ``LibA/1.0``. Then the consumers, when they upgrade to the new one, they will upgrade to ``liba/2.0``. It will only fail in Windows, if trying to use both versions, but cleaning the cache will solve it. The casing shouldn't be a bigger problem than upgrading to a new major version, is it?

dean0x7d:
I wholeheartedly agree with lowercase references and I wish it was this way from the start, as I'm sure you do too. However, CI migration is going to be a major roadblock for us. And it's the kind of roadblock that will not be possible to fix by a migration grace period. It can only be fixed with migration tools built into Conan. Let me try to explain since we've had this exact same roadblock already while trying (and failing) to transition from `OpenSSL` to `openssl`.

When Conan Center renamed `OpenSSL` to `openssl`, we tried to transition. It's pretty easy on local systems: you get the error message, you manually delete the old cache entry, all done. But on CI it turns into a nightmare. For good CI performance, we keep a node-local Conan cache (always fetching from Artifactory/Conan Center is too slow for the number and size of dependencies that we have). The Windows CI will fail on the `Rename OpenSSL to openssl` commit because Conan raises the case-conflict-in-cache error. So what can we do at this point?

* **Option 1**: Manually clear the CI conan cache at this point and make this a hard migration point. -- This was a no-go for us since we need to keep pre-rename branches and projects running and manually clearing the CI cache all the time is a nightmare.

* **Option 2**: Find a way to automatically clear and re-fill the offending cache entry on conflicts. -- This would be ideal. If a branch/project requires `openssl` but `OpenSSL` is in the cache, the CI would automatically delete the old entry and download the new one. If an older branch/project needs the old `OpenSSL`, it would do the opposite. There's a small extra cost to invalidating and re-downloading dependencies for these transitions but it would be perfectly fine, especially since it would be automatic. However, we were unable to find a robust way to do this (one that would not involve parsing Conan's error message output -- something that may break with newer versions).

* **Option 3**: Download the latest `openssl`, rename it to `OpenSSL` and upload it to our Artifactory. -- This is a really dumb solution but it's completely non-disruptive (unlike option 1) and it was feasible to do quickly (unlike option 2). Of course, this is a no-go for Conan 2.0 since uppercase references would be completely forbidden but I'm just describing what we ended up doing for this existing Conan 1.x issue.

I can see that we'll run into the same exact scenario with `OpenSSL` for the Conan 2.0 transition, but option 3 will be off the table. Option 2 would be great if it was a builtin setting in Conan: "If there's a case conflict in the cache, just delete the old cache entry (don't raise an error)". This would definitely resolve the CI migration issues in our setup. I understand it would not be universal, e.g. it does not resolve the issue of having both `openssl` and `OpenSSL` on different parts of the graph within the same project, or the issue of serving multiple projects concurrently (a CI node does multiple projects but only one at a time). Alternatively, providing a hook so that we can detect and take some (user-defined) action if this conflict happens would also work for us (although much more complicated than the simple option mentioned before).

Again, I'm writing all of this mainly from a CI migration point of view, not project migration. The grace period will work for project migration since there's manual work involved that will take time. For us, the CI migration is a technical issue, not one of effort. We need some way for Conan to deal with cache conflicts: a simple option to invalidate old cache entries would work for us.

DoDoENT:
@dean0x7d, here is my super-secret master plan for renaming all my packages into all-lowercase. It's based your **Option 3**, but with some spice added.

First, I'll rename my `ZLib/<version>@company/stable` into `z-lib/<version>@company/stable` and use the [provides attribute](https://docs.conan.io/en/latest/reference/conanfile/attributes.html#provides) to inform conan that `z-lib` provides both `ZLib` and `zlib` (the name in conan-center).
I'll do the same for `OpenCV` - the new name will be `open-cv`.

Then, I will gradually update my packages to use `z-lib` and `open-cv` instead of `ZLib` and `OpenCV` and since the name change does not only change in case, the Conan cache will not have any problems with it - I can even start today with Conan 1.28+ (Conan v1.28 is required for `provides` attribute to work).

Then, in a distant future when no package will use either `ZLib` or `OpenCV`, I can perform the same transition once more, but this time using "official" names: `zlib` and `opencv`.

You can do the same with `OpenSSL` (`open-ssl` looks OK to me 😛 ).

Furthermore, if [this feature request](https://github.com/conan-io/conan/issues/8275) gets implemented, the transition could be even faster, as most downstream packages could override their dependencies to use `z-lib` and `open-cv` (and later `zlib` and `opencv`) instead of `ZLib` and `OpenCV` without waiting for all our packages to get updated.

memsharded:
@DoDoENT @dean0x7d

I wanted to share the current thinking about the Conan cache design. As we are going to move to always using revisions, it is likely that a redesign of the cache to account for multiple revisions is necessary. Putting revisions in the the path might not be possible, too long paths, even for Linux systems, so we might need some kind of abstraction over the paths. We don't know yet how it will look like, but if this happens, we might actually have an automatic mechanism that will allow to have different casing packages even in Windows systems. Basically because the folders ``OpenSSL/1.0/user/channel/revision/package_id/package_revision`` and ``openssl/1.0/user/channel/revision/package_id/package_revision`` need to be abstracted in a hash or something. 

This might allow to have non-conflicting packages in the cache, and help with the migration, but note it will not affect how packages are resolved, they will still be different packages and case sensitiveness will remain.

DoDoENT:
> This might allow to have non-conflicting packages in the cache, and help with the migration, but note it will not affect how packages are resolved, they will still be different packages and case sensitiveness will remain.

Well, in that case, I see no reason to enforce lower case package names for internal company packages. I still agree that this should be the case for packages in conan-center (primarily for reasons you mentioned in the proposal), but for packages in the controlled environment in the company's private Artifactory server, I see no major benefits of banning mixed-case package names...

memsharded:
> Well, in that case, I see no reason to enforce lower case package names for internal company packages. I still agree that this should be the case for packages in conan-center (primarily for reasons you mentioned in the proposal), but for packages in the controlled environment in the company's private Artifactory server, I see no major benefits of banning mixed-case package names...

It is a current major source of conflicts between proprietary packages and ConanCenter packages. The number of users using packages from ConanCenter directly (via Artifactory, but without rebuilding nor changing anything) into their builds is growing, and if they have their own copies of packages like ``OpenSSL``, then the conflict when they include something from ConanCenter that depends on ``openssl`` package, it is a nightmare that cannot be addressed.

dean0x7d:
@memsharded That sounds very good to me! It would definitely take care of our migration issues if both package cases can happily live in the same cache. And it sounds like it would also do away with OS-specific filesystem quirks, which would be a nice bonus.

monsdar:
Sorry for being late to the party. I'm with @Daniel-Roberts-Bose I'm afraid... We're having more than 600 packages by now and all of them are named with CamelCase for better readability.

If there'd be a way to easily migrate that would help, but still could be a major effort. Not speaking of backwards compatibility etc.

I know this issue has been discussed thoroughly in the past and also within this PR. I've been following the discussions loosely and see that there is no "best way" to follow. Nonetheless I'd like to bring up the idea of not relying on the filesystem for managing the cache once more.

-> Have we already ruled out storing this kind of information within a local database? All data regarding a package could be stored in a SQLite DB (name of the package, directory where everything is stored, ...). The directory itself could be anything then, even differ from OS to OS (use package name within the path on Linux, have short pathes on Windows by default, ...). The information where everything is stored could then be available by using `conan inspect` for example. Thirdparty-scripts could then rely on the output of `conan inspect` instead of relying on an implementation detail like the filesystem structure. This increases maintainability and extendability.

DoDoENT:
> Have we already ruled out storing this kind of information within a local database?

I'm afraid such a thing wouldn't work for us, as we have custom scripts that operate on Conan cache and depend on its layout (we plan to update the scripts for Conan v2.0, but I'm not sure if that would work if the database were used).

For example, we have a script that scans the Conan cache for all packages that contain the `build` folder and builds a list of those packages. That list can then be used to on-demand upload the packages to Artifactory so they won't have to be built in the next CI build, possibly running on a different node.

memsharded:
> -> Have we already ruled out storing this kind of information within a local database? All data regarding a package could be stored in a SQLite DB (name of the package, directory where everything is stored, ...). The directory itself could be anything then, even differ from OS to OS (use package name within the path on Linux, have short pathes on Windows by default, ...). The information where everything is stored could then be available by using conan inspect for example. Thirdparty-scripts could then rely on the output of conan inspect instead of relying on an implementation detail like the filesystem structure. This increases maintainability and extendability.

@monsdar that could be close to our current way of thinking about the Conan cache. To store multiple revisions in the cache, something like what you propose would be necessary. We understand it is a change in the cache layout, an abstraction, so to help with this new cache, new commands (like ``conan cache folder build``, for example), might be a thing. 

@DoDoENT please note that many of the changes that will come with Conan 2.0, and some of the latest improvements in Conan should really help with this kind of operations. For example, uploading in Conan 2.0, with always revisions on, and package immutabilities, would be much faster: no need to compress anything, or to check manifests. Just checking the revisions will tell which things need to be uploaded and what things are already in the server. Also the ``lockfiles`` provide information of what has been changed, which are the things that needs to be uploaded. That is the information that should be used in order to do an upload. It is possible that helpers that extract that directly to pipe it into the ``conan upload`` command will be implemented.

DoDoENT:
> @DoDoENT please note that many of the changes that will come with Conan 2.0, and some of the latest improvements in Conan should really help with this kind of operations. For example, uploading in Conan 2.0, with always revisions on, and package immutabilities, would be much faster: no need to compress anything, or to check manifests. Just checking the revisions will tell which things need to be uploaded and what things are already in the server. Also the lockfiles provide information of what has been changed, which are the things that needs to be uploaded. That is the information that should be used in order to do an upload. It is possible that helpers that extract that directly to pipe it into the conan upload command will be implemented.

Well, let me describe our case so we'll see if this would be possible in the way you suggest.

First, we have two real and one virtual Conan repository in Artifactory. One repository (named `conan-local`) is the main repository containing stable packages that should be used downstream. Those are built only using configurations that we need. The other (named `conan-temporary`) is used as "binary cache", i.e. all developers have both read and write access to it and can deploy arbitrary packages to it, as well as delete anything from it. It serves mostly to avoid long build times when dependency does not exist or when some local version of a package needs to be tested on CI. Finally, there is a virtual repository (named `conan-virtual`), which is backed by the mentioned two repositories for reading and by `conan-temporary` for writing. This is a default remote for all developers - it will contain both binaries built by the CI from `conan-local` (i.e. "official binaries") and binaries built by either developers or CI (see below for workflow) in `conan-temporary`.

Second, there are two types of workflows for our CI system:

The first workflow is used when building a `stable` branch, i.e. when making a release. Here the CI is allowed to use only `conan-local` repository and builds all packages from the repository. After all packages from the repository have been built, it uses the script I described in the earlier comment to detect all packages that have been built, either directly or by dependency and uploads them to the `conan-local` repository. This is primarily used instead of `conan upload '*'` as it's much faster than simply letting Conan do the detection of what needs to be uploaded. I guess with revisions enabled by default and new cache layout Conan could be much faster in doing that (we'll need at least a factor of 100 here).

The second workflow is used when building pull requests and feature branches. Here the CI invokes the build script (usually cmake, but not always directly (could be via gradle for android or npm for emscripten) to build the code and run the tests. In order for next build of the same PR/branch to be faster, all Conan package binaries that were built by dependency are uploaded to `conan-virtual` repository (effectively into `conan-temporary`, but unlike proper upload it may upload only missing binaries for already existing recipe in `conan-local`). Thus, if next build ends-up on a different node, Conan does not have to build the same binaries again. On the other hand, every developer can always wipe the entire `conan-temporary` as it may also contain binaries that are built against local exports of conan packages that may not yet be stable.
I guess that `conan upload '*'` may work for us here as well if it gets at least 100 times faster as it currently is and will work with virtual conan repositories.

monsdar:
>  that could be close to our current way of thinking about the Conan cache. To store multiple revisions in the cache, something like what you propose would be necessary. We understand it is a change in the cache layout, an abstraction, so to help with this new cache, new commands (like conan cache folder build, for example), might be a thing.

This sounds very good. With such a system in place I think there is no need to force lowercase package names? It would be easy to store the pathes where everything resides in regards to the stored package name.

memsharded:
@monsdar 

> This sounds very good. With such a system in place I think there is no need to force lowercase package names? It would be easy to store the pathes where everything resides in regards to the stored package name.

Even in this case, there are other problems, like commented in https://github.com/conan-io/tribe/pull/17#issuecomment-752681917, and in the proposal itself: the confusions and problems that comes from having the same library/code packaged with different casing, that are considered as different packages, creating non-solvable conflicts. The fact that the filesystem can store them simultaneously doesn't address them. 

We are proposing to leave a opt-out for users that have packages with CamelCase, but it would be very beneficial for the whole community to converge as much as possible to lowercase packages.

memsharded:
@DoDoENT 

Yes, I think that will be definitely possible. If the improvements in the upload checks are not enough, it is extremely likely that we will use the output of the lockfiles, to define the things that need to be uploaded. That will be integrated, and very efficient, as the lockfiles store that information of what has been built.

Regarding your flow, I would recommend to spin yet another Artifactory local repository. The fact that CI binaries are being uploaded to a repo that developers can wipe things is a bit concerning IMHO. I would do:

- One "stable" repository, for "stable" packages and build (write permissions only by CI)
- One "ci-builds" repository, for "branches" packages and build (write permissions only by CI)
- One "devs" repository, this the users can use to upload packages (write permissions by users, CI will not use this one at all)

Eventual binary promotions might happen only in "ci-build" => "stable", but not from "devs" repository.

DoDoENT:
> Regarding your flow, I would recommend to spin yet another Artifactory local repository. The fact that CI binaries are being uploaded to a repo that developers can wipe things is a bit concerning IMHO. I would do:

It sounds concerning, but in fact, it's not. We don't do promotions, so our two repositories are in fact the "stable" and "devs" when using your terminology. The CI can build and deploy into "devs" just to save some compile times to devs that wish to share their packages with other devs prior making an official release (the servers can compile the code much faster than developer machines). Now, since there can be multiple iterations of the dev package before making a stable one, developers are allowed to wipe this repository in order to avoid raising package version numbers if they need to make a change that would require recompilation.

For example, a developer working on `liba` wants to share its new version v2.0.0 with developers working on `libc` in order to test it before making `liba/2.0.0` stable. However, `libc` uses `liba` transitively via `libb/1.0.0@company/stable`, which is available in stable Conan repository. After publishing the `liba/2.0.0@user/testing` into "conan-temporary" (i.e. "devs") repository, a developer working on `libc` overrides the dependency with `liba/2.0.0@user/testing` and ask the CI to build all binaries (including `libb/1.0.0@company/stable`) and deploy them to "conan-temporary" so other developers working on `libc` don't have to wait for `libb` binaries to compile on their devices (this can take a lot of time on laptops, but not so much on servers). After performing tests, developers working on `libc` give feedback to developers working on `liba` so they can update their changed API for v2.0.0 to be even better.

Now, when a second iteration begins, developers working on `liba` make a new change that [could be breaking](https://docs.conan.io/en/latest/creating_packages/define_abi_compatibility.html#dependency-issues), but don't want to raise liba's version number to v3.0.0 just to ensure that `libb` will be recompiled. So, developer wipes the "conan-temporary" repository (which contained binaries for `liba/2.0.0` and `libb/1.0.0` build against `liba/2.0.0`) and the entire iteration, described above, can start anew.

This actually works around the behaviour of conan's [`semver_direct_mode` and `semver_modes` version schemas](https://docs.conan.io/en/latest/creating_packages/define_abi_compatibility.html#versioning-schema) that ignore the user/channel. AFAIK, there is no version schema that will use `semver` mode, but only for the same user/channel combination, i.e. dependency on `liba/2.0.0@company/stable` and `liba/2.1.0@company/stable` would result with same package ID, but different from package ID when dependency is `liba/2.0.0@user/testing` or `liba/2.1.0@user/testing`. 

Or am I missing something?

memsharded:
This proposal is going to be merged with modifications from https://github.com/conan-io/tribe/pull/17/commits/1e794f7aaff430292b6397753b529381ba432083. As a summary:

- The major concern seems the impossibility of the Conan cache to store packages with different casing, specially on Windows. As the cache needs to be upgraded to be able to store multiple revisions, Conan 2.0 will implement such functionality regarding the package casing for all OSs, including Windows. This will allow this behavior, so far impossible in Conan 1.X, and that will allow for an easier migration.
- Packages not in lowercase will have a temporary opt-out. This temporary opt-out will be maintained at least 6 months.

With the above measurements, migrating would be 100% functionally equivalent to a major version bump of packages, which is a known process that can be migrated incrementally without major disruptions.

Da-LiFe:
I just found in the conan 1.x docs under [name](https://docs.conan.io/en/latest/reference/conanfile/attributes.html#name) and [version](https://docs.conan.io/en/latest/reference/conanfile/attributes.html#version) these examples:

- `<PkgName>/version@user/channel`
- `PkgName/<version>@user/channel`

I'm afraid that having uppercase examples in the docs might cause people to apdopt this as a kind of style guide for their new packages. Although these are correct examples for conan 1.x, maybe its worth updating the conan 1.x docs already regarding this proposal, i.e. not using uppercase in the reference of examples?

memsharded:
@Da-LiFe agree, done https://github.com/conan-io/docs/pull/2013, please report in https://github.com/conan-io/docs if you find more occurrences of this. Thanks!

monsdar:
> * The major concern seems the impossibility of the Conan cache to store packages with different casing, specially on Windows. As the cache needs to be upgraded to be able to store multiple revisions, Conan 2.0 will implement such functionality regarding the package casing for all OSs, including Windows. This will allow this behavior, so far impossible in Conan 1.X, and that will allow for an easier migration.
> * Packages not in lowercase will have a temporary opt-out. This temporary opt-out will be maintained at least 6 months.

Just to make sure I understand correctly:
* Conan Cache will be updated to handle multiple revisions. Along with that there will be an update to support mixed casing.
* Even though the Cache will support mixed casing then, this will still be an opt-out and support removed after 6 months.

I hoped that with the Cache supporting mixed casing there will be no need to force lowercase packagenames. But if that's the outcome then I will start planning the migration.

memsharded:
Hi @monsdar 

Yes, that is correct.

> I hoped that with the Cache supporting mixed casing there will be no need to force lowercase packagenames. But if that's the outcome then I will start planning the migration.

The problem is that even if the cache will be able to support mixed case package names, it is still an important problem. This is not only about the cache limitations. This is already supported in Linux, but still biting Linux-only users. Allowing users to create packages with UpperCase, like ``OpenSSL`` is creating a lot of issues and unsolvable problems when they try to use another package recipe for example from ConanCenter, that will transitively depend on ``openssl``. At some point, it is needed that the whole community moves to the same convention of package names to avoid this situation, or these unsolvable conflicts will keep happening forever. It is fair to say that this is the "users responsibility" and that they should know what they are doing, but the reality is that we keep getting users complaining to us, not happy that this is happening to them.

DoDoENT:
> Allowing users to create packages with UpperCase, like OpenSSL is creating a lot of issues and unsolvable problems when they try to use another package recipe for example from ConanCenter, that will transitively depend on openssl

That would not necessarily resolve the fundamental issue. Companies would still be able to create internal packages like `open-ssl` or `open_cv` which would cause the very same problems as currently `OpenSSL` and `OpenCV` cause.

I'm afraid that this needs to be the "users responsibility" and here I am with @monsdar.

memsharded:
> That would not necessarily resolve the fundamental issue. Companies would still be able to create internal packages like open-ssl or open_cv which would cause the very same problems as currently OpenSSL and OpenCV cause.

It will not completely remove the fundamental conceptual issue, but at least it would greatly mitigate it. Absolutely all of the support tickets we have received so far related to this issue are due to package names different casing, not different variants like ``open_cv``. Of course it can happen, but so far it is not a real problem in practice. While very uncomfortable situations the maintainers need to deal with, and cause stress, frustration and burnout in both sides, is a very real problem, that could be avoided at least in >95% of the cases. I am 100% with you too, it should be "users responsibility". But "users responsibility" sounds great until most of the users hitting this issue disagree and they think that it is Conan and the maintainers responsibility.

DoDoENT:
> I am 100% with you too, it should be "users responsibility". But "users responsibility" sounds great until most of the users hitting this issue disagree and they think that it is Conan and the maintainers responsibility.

Maybe this can be addressed with better documentation. Most users of Conan are C++ developers, and C++ developers are no strangers to "user responsibility" - just think about variable initialization (Should I use parenthesis or braces? Why is sometimes irrelevant and sometimes not (very analogous to the Conan problem you identified here)? ... )

DoDoENT:
>  Absolutely all of the support tickets we have received so far related to this issue are due to package names different casing, not different variants like open_cv.

Were those tickets related only to renames in conan-center (like `OpenSSL -> openssl` and `OpenCV -> opencv`) or also because of different casing between local company's package name vs open-source package name?

I'm asking because all GitHub issues I've seen were related to `OpenSSL -> openssl` rename in the conan-center.

memsharded:
This has been implemented and will be in 2.0-alpha2: https://github.com/conan-io/conan/milestone/180. The opt-out still not implemented, but will be soon.

PR #24, opened 2021-06-04
Do not automatically create a default profile

memsharded:
The automatically created ``default`` profile has had some relevant issues in the last year, as the default definition of ``libcxx=libstdc++`` that become obsolete some time after it was created, but couldn't be changed without breaking, and what is worse, produced linking errors where not easy to understand by users.

This proposal is about making the creating of the default profile explicit: When the "default" profile in the ``.conan/profiles`` folder, does not exist, and is required by some Conan command, raise an error indicating that the profile doesn't exist, and telling the user to explicitly create it.

There will be one Conan command, like the current ``conan profile new default --detect`` (hopefully with better UI) that will perform the autodetection. The result of the autodetection will not be considered stable.

---

### DIFF
--- a/design/024-explicit_default_profile_creation.md
+++ b/design/024-explicit_default_profile_creation.md
@@ -1,0 +1,62 @@
+# Proposal: Do not automatically create a default profile if not existing.
+
+
+| **Status**        |                                                   |
+|:------------------|:--------------------------------------------------|
+| **RFC #**         | [024](https://github.com/conan-io/tribe/pull/24)  |
+| **Submitted**     | 2021-06-02                                        |
+| **Tribe votes**   |                                                   |
+
+
+## Summary
+
+When the "default" profile in the ``.conan/profiles`` folder, does not exist, and is required by some Conan command, raise
+an error indicating that the profile doesn't exist, and telling the user to explicitly create it.
+
+There will be one Conan command, like the current ``conan profile new default --detect`` (hopefully with
+better UI) that will perform the autodetection. The result of the autodetection will not be considered stable.
+
+
+## Motivation
+
+In one of the initial Conan versions, the default ``libcxx=libstdc++`` was defined, also for ``gcc>=5`` compilers. At that time, this was the default standard library in many mainstream Linux distros **even
+if you upgraded your compiler version to ``gcc>5``. It was a good default, provided the largest binary
+compatibility, etc.
+
+With the upgrade of the Linux distros and compiler versions, ``libcxx=libstdc++11`` became the default,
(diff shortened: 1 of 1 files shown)


mathbunnyru:
That's a great change for beginners - it is so awkward to see libstdc++ (instead of libstdc++11) being used by default in modern OS, with modern toolchain and etc.
And the error afterward was quite confusing to me (I had to google, I remember it).

memsharded:
This might be a bit counter-intuitive for some people, thinking that the more automatic and transparent, the better for beginners, but I agree with you: this will be better for beginners, because it will help to understand an important and relevant information that affects later commands results, the earlier beginners are aware of their default profile, the earlier potential mismatches can be caught.

uboot:
Should there be a flag for Conan commands to silently create a default profile if there is none? This would be the same behavior as now but opt-in only.

The use-case I have in mind is the following: the package build environment is provided as a docker container which has build tools and Conan installed in it. Inside this container the auto-generated default profile is stable because the environment and the Conan version are fixed. In other words, the container carries an implicit default profile. In such a case one has to explicitly generate a default profile before e.g. creating a package. This is not a big deal if done manually but means an extra step to maintain in case of a CI setup. Alternatively, one could of course install a default profile into the container image.

memsharded:
@uboot 

What is the difference between a flag in Conan commands and a explicit command to create the default profile? The second sounds much cleaner than adding a new flag to every Conan command.

For the Docker case, I totally see the initialization of a default profile inside the image, which could be done by explicit ``conan config install`` (might be necessary for many other config items), or by a ``conan profile detect default`` command. In either case, it seems better to provide it initialized in the image, so users of the image will have it there and not have to worry to remember to do anything else.

sztomi:
Even though this does not affect us very closely, I like the direction towards less magic and more explicit behavior.

rockdreamer:
We will be impacted by this as we use the default profile as a root in a profile hierarchy. Requiring an additional command is not a huge burden.
Keeping the autodetection ability IMHO is fundamental for learners.

Thank you for communicating these changes this way, I can plan for the change. :)

memsharded:
> Keeping the autodetection ability IMHO is fundamental for learners.

Just to clarify, the autodetection of the profile will remain, conan will still be able to create a profile with the "best guess" based on what is installed on the machine. The proposal is about this doesn't happen automatically when such profile doesn't exist.

memsharded:
It seems this proposal has good support, with no one opposing to it. Less magic, more explicit = good.

Lets merge it and start implementing it!

PR #25, opened 2021-06-04
Make alias requirements explicit, similar to version range requires.

memsharded:
The ``alias`` requirements are problematic, they have accumulated many issues in the last years. The original root cause of these issues is the lack of information and impossibility to know that an alias requirement is actually an alias, until it is fully fetched and resolved. While the version ranges have the ``pkg/[>1.0]`` brackets differentiator, the alias have nothing.

So the proposal is to make "alias" requirements explicit, in the form ``requires = "pkg/(alias)@user/channel"``, where the ``alias`` can be any user string, as "latest".

Alias packages will remain named, implemented and managed the same. For a requires like ``requires = "pkg/(latest)@user/channel"``, there must exist a ``pkg/latest@user/channel`` package that will contain a single ``alias`` attribute pointing to the aliased real package. This doesn't change with respect the Conan 1.X status.

---

### DIFF
--- a/design/025-make_alias_explicit.md
+++ b/design/025-make_alias_explicit.md
@@ -1,0 +1,57 @@
+# Proposal: Make alias requires always explicit.
+
+
+| **Status**        |                                                   |
+|:------------------|:--------------------------------------------------|
+| **RFC #**         | [025](https://github.com/conan-io/tribe/pull/25)  |
+| **Submitted**     | 2021-06-02                                        |
+| **Tribe votes**   |                                                   |
+
+
+
+## Summary
+
+Make "alias" requirements explicit, in the form ``requires = "pkg/(alias)@user/channel"``, where the
+``alias`` can be any user string, as "latest".
+
+Alias packages will remain named, implemented and managed the same. For a requires like ``requires = "pkg/(latest)@user/channel"``, there must exist a ``pkg/latest@user/channel`` package that will contain a single
+``alias`` attribute pointing to the aliased real package. This doesn't change with respect the Conan 1.X status.
+
+
+## Motivation
+
+Alias is a complex and fragile feature, that has had many bug reports and issues in the last years. Most
+of the source of this complexity and fragility is that there is no indication whatsoever in a ``require``
+that such a requirement is not a real package, but kind of a symlink to the actual package that will be used.
+So when a package declares something like ``requires = "pkg/latest@user/channel"``, it will not depend on anything called "latest", but on something like ``pkg/3.1.2@user/channel``.
+
+This makes the dependency resolution challenging, because any requirement could be not an actual dependency,
+but an alias to the real dependency. When there are complex graphs, with diamonds and possible conflicts,
+then detecting the conflicts becomes challenging, because 2 different things like ``requires = "pkg/latest@user/channel"`` and ``requires = "pkg/1.0@user/channel"`` could be aliases to the same ``pkg/1.0.0@user/channel`` package. Not only the business logic of resolving those, but currently there is a good amount of code in Conan oriented to cache the alias resolution, as they are slow to resolve, and it is
+common that  projects will use aliases in different packages, so this is necessary to keep good performance.
+
+In the core of the issue, is the before mentioned lack of information in the requirement itself. Contrary
+to the version ranges requirements, which define ``requires = "pkg/[>1.0]@user/channel"`` and the brackets
+indicate that such dependency is not a real package, but it could resolve to something like 1.1, there is nothing in "alias" requirements that provide such a hint.
+
+
+## Proposal
+
+Introduce an alternative notation, like ``requires = "pkg/(alias)@user/channel"`` to indicate that such
+requirement is not a real package, but an alias that will be resolved to something different, like ``pkg/1.0.2@user/channel``. This will establish a parallelism between version ranges and alias, make explicit to package writers and most importantly, readers, the intention to depend on a moving target that will be defined by an alias. It will also allow to simplify the dependency resolution algorithm, making it less fragile and probably faster.
+
+
+## Detailed Design
+
+While the ``requires`` syntax will allow ``()`` characters, the package names and versions will remain with the same restrictions, real package versions will not be allowed to contain the ``()`` characters, in the same way they do not allow the ``[]`` brackets.
+
+Current resolution features will be maintained:
+
+- It is possible to have transitive alias (one alias package aliasing other ones). The only requirement is that the alias definition should also be explicit: ``alias = "pkg/(otheralias)@user/channel"``
+- Alias never appear in the final dependency graph, they are discarded in the graph evaluation process.
+- Alias do not have settings, options, or any configuration. They cannot have python_requires, they are a recipe with 1 single class attribute ``alias``.
+
+
+## Migration plan
+
+The introduction of the ``()`` syntax might be introduced in Conan 1.X requirements, with no effect rather than filtering them out and ignoring them (they will still be resolved correctly, up to the limitations of Conan 1.X alias dependency resolution), keeping Conan 1.X working and allow the smooth transition to Conan 2.0.


puetzk:
Why a new syntax `pkg/(alias)@user/channel` instead of just extending the existing syntax `pkg/[alias]@user/channel`? The current semver version range expressions have to start with a digit or an operator, so if alias names had to start with an alphabetical character (or even just require them to be identifiers), this would be unambiguous. And they seem like very related concepts; both put off resolving a specific version until they have looked at existing packages in the cache.

petermbauer:
This raises the questions:
- what happens when an alias package is used with the regular version spec (no parentheses)?
- what happens when a regular package is used with the parentheses syntax?

memsharded:
@puetzk The problem is that ``pkg/[1.0]@user/channel`` is both valid semver version-range expression, and there are users that are using a version number to alias another version number, like ``1.0`` is an alias for ``1.0.123``. Then, we are again at the same issue of lack of enough information, in one case it is mandatory that it will resolve to that version or nothing, and in the other it must resolve to something different. This is the kind of behavior and user expectations that we would like to make more evident.

memsharded:
@petermbauer good points.

> what happens when an alias package is used with the regular version spec (no parentheses)?

It will resolve as a normal package, and as package recipes that define ``alias`` are basically empty, it will surely fail later. We could make it fail more explicitly, if that helps with the UX.

> what happens when a regular package is used with the parentheses syntax?

It will explicitly fail, as it will not find the ``alias`` attribute in the recipe.

Does it make sense? If yes, I think we can add them to the doc.

petermbauer:
> Does it make sense? If yes, I think we can add them to the doc.

👌

Da-LiFe:
Currently, we have implemented this release strategy:

- release branches become unstable releases, e.g. `release/1.2.0` => `<name>/1.2.0@<user>/unstable`
- tags become stable releases, e.g. `1.2.0` => `<name>/1.2.0@<user>/stable`

As the stable release superseeds the unstable one, we are removing the packages from artifactory after some time to save disk space. To make that process transparent for package users, we currently deploy also an alias during the stable release:

 `<name>/1.2.0@<user>/unstable` => `<name>/1.2.0@<user>/stable`

However, having the proposed new semantics, a conanfile that requires `<name>/1.2.0@<user>/unstable` would fail after the package is replaced by an alias, right? Could you think of a migration path for our release strategy? The tricky part is that we are using aliases like symlinks so the users actually should not need to know the fact that they are requiring an alias instead of a real package.

puetzk:
> there are users that are using a version number to alias another version number, like 1.0 is an alias for 1.0.123

Sorry I was unclear, I was counter-proposing that (since this is a breaking change anyway), this might be disallowed in 2.0, requiring alias names to start with something other than a digit. Mostly because then we'd have only one kind of syntax for an indirect version lookups that must be resolved by consulting the cache/remotes to pick a specific "real" version.

But that would definitely not work with a use-case like what @Da-LiFe describes. Although that seems like a strange practice anyway... artifactory does hash-based deduplication, so removing the "real" package to replace it with an alias isn't saving any disk space, it's in fact spending just a tiny bit more (to store the alias, whereas the other URL referenced the same blobs).

Daniel-Roberts-Bose:
>  While the version ranges have the pkg/[>1.0] brackets differentiator, the alias have nothing

This statement confuses me. We have been using the version range syntax on alias' for years now and it has been working fine. Can you provide some clarity on this?

A reference such as `package/[>=0.0.1, include_prerelease=True]@user/alias_to_master` has been working fine for us for a very long time

memsharded:
@Daniel-Roberts-Bose 

> A reference such as package/[>=0.0.1, include_prerelease=True]@user/alias_to_master has been working fine for us for a very long time

Do you mean that you have a version-range that resolves to a versioned reference that in turn happens to be an alias to still indirect to another package?

Daniel-Roberts-Bose:
@memsharded 
> Do you mean that you have a version-range that resolves to a versioned reference that in turn happens to be an alias to still indirect to another package?

My workflow is something like this
`conan create conanfile.py package/1.0.0@user/channel` then
`conan alias package/1.0.0@user/channel  package/1.0.0@user/alias_to_channel`

then using `package/[>=0.0.1]@user/alias_to_channel` in my requirements, which resolves and works fine

memsharded:
@Da-LiFe 

I didn't fully understand your flow, specifically that transition that requires an alias. If you have your recipes with ``requires = "pkg/1.2.0@user/unstable"``, and those recipes are now tagged, you have your "production", released code pointing to "unstable" packages, and bypassing that with an alias. When you modify the unstable (if you do) in production packages and make them point to "stable" packages, then you are generating new revisions (and thus new binaries too for the new revision), that no longer are the ones built when the tag was done, but later. 

I think this is strongly connected to the "immutability" proposal here in the tribe in https://github.com/conan-io/tribe/pull/16. Following those principles there the flow would be more like:

- You have a "unstable" server repo. Your ``release/1.2.0`` branches packages go there. They will always use the same, constant ``user/channel``.
- When you build your release branches, or PRs, you can add this server repo, to resolve packages there.
- When the tag is done, if the tag comes from a merge that is basically a fast-forward from a release branch, then package binaries can be directly promoted. No need to rebuild them (actually a devops principle say that they shouldn't be rebuilt, they were already created, tested and validated in ``release/1.2.0``). 
- The promotion works by copying package binaries from ``unstable`` server repo to the ``stable`` server repo (This is very efficient in Artifactory, for example, because of artifacts de-duplication).

Please note that under many ``package_id_modes`` and most likely the default ones that will be used in Conan 2.0, changing one of the dependencies channel will invalidate and produce new ``package_id`` in the consumers, so the "transparent" replacement that an alias could do would not work to keep using same binaries, and will still require a rebuild.

memsharded:
@Daniel-Roberts-Bose 

Oh I see, it seems you are aliasing over the channel, which apparently was not an intended use case, the documentation only shows aliases over the ``version`` field, like ``latest => 1.0.0``, not for aliasing the same version but on other channel.

I would say that the above comment also applies here. For Conan 2.0, we probably shouldn't use anymore the channels as a versioning mechanism, and let it be just a general label, same as the "user" part. Channels shouldn't be used for flow control or dependency resolution logic, and using different servers and following the immutability principle of packages should be the approach to follow.

ytimenkov:
I like removing alias from the name because it breaks things later: I tried to provide a "compatibility" name when we did some reorganization and renamed a package, but it didn't work: even if "new" package was fetched via an alias to the "old" one, all targets in `conanbuildinfo` were still referring to a wrong name.

Da-LiFe:
> Although that seems like a strange practice anyway... artifactory does hash-based deduplication, so removing the "real" package to replace it with an alias isn't saving any disk space, it's in fact spending just a tiny bit more (to store the alias, whereas the other URL referenced the same blobs).

Oh, we weren't aware of this fact. Thanks for pointing that out. It indeed makes no sence the create aliases for this use case then!

memsharded:
> Oh, we weren't aware of this fact. Thanks for pointing that out. It indeed makes no sence the create aliases for this use case then!

This is important, because all the proposed "promotion" based flows are also very efficient and this is another reason they are recommended. As Artifactory de-duplicates, promotions, copying packages from one repo to another doesn't duplicate the artifacts, and it is a very fast operation. It is also way more atomic that any operation that involves an actual upload.

Da-LiFe:
> For Conan 2.0, we probably shouldn't use anymore the channels as a versioning mechanism, and let it be just a general label, same as the "user" part. Channels shouldn't be used for flow control or dependency resolution logic, and using different servers and following the immutability principle of packages should be the approach to follow.

I'd like to discuss this different view on channels in Conan 2.0 further, but it's off-topic to this proposal. Should I open an issue in the conan repo, in the tribe repo or were would be a good location for this?

Daniel-Roberts-Bose:
@memsharded 
> I would say that the above comment also applies here. For Conan 2.0, we probably shouldn't use anymore the channels as a versioning mechanism, and let it be just a general label, same as the "user" part. Channels shouldn't be used for flow control or dependency resolution logic, and using different servers and following the immutability principle of packages should be the approach to follow.

I need to experiment further with this as I haven't tired it before however, doesn't this really limit our ability to promote code to say, a stable environment or something? In my eyes the whole point of promoted a package to something like stable is to simply "label" it as you said with a new channel. 

So far as I can tell, if you use the version to alias wont that mean that only one stable package, for a given package name, can exist in artifactory at a time? The layout in artifactory is user -> package name -> version -> channel so this would only allow for one stable version at a time which seems limiting. Its highly possible I am mis understanding some element of this but can you shed some insight on this?

memsharded:
> I'd like to discuss this different view on channels in Conan 2.0 further, but it's off-topic to this proposal. Should I open an issue in the conan repo, in the tribe repo or were would be a good location for this?

@Da-LiFe yes, I think an issue in the Conan repo would be the best, to understand a bit better the scope (and better using an issue and in this repo we only have PRs at the moment). Just to clarify, this view on channel is not something completely new for Conan 2.0, we have already been recommending this for some time, and the usage of multiple repositories is also documented in the newest sections of the docs, like CI flows: https://docs.conan.io/en/latest/versioning/lockfiles/ci.html#repositories.  So yes, please, open an issue in the Conan repo and we will talk there.

memsharded:
@Daniel-Roberts-Bose 

> I need to experiment further with this as I haven't tired it before however, doesn't this really limit our ability to promote code to say, a stable environment or something? In my eyes the whole point of promoted a package to something like stable is to simply "label" it as you said with a new channel.

Actually it seems that it achieves the goal better, not limit it:

- When some packages are working fine, many times it is not necessary to rebuild them, after some quality checks and testing they can be validated or promoted to a new stage. Requiring re-builds of already validated binaries introduces risks (would require to re-validate, in theory)
- Promotions changing channels actually breaks data consistency. If you are building app/1.0@user/testing -> pkga/1.0@user/testing -> pkgb/1.0@user/testing, and you want to promote it to stable, it will change the package names, but internally, all conaninfo.txt files will remain with the old user/testing, even if the packages are user/stable. That means that even package_ids could be broken, or at least not correct.
- Lockfiles will keep working, as they lock full references, but can resolve against different remotes. Changing user/channel is a new graph, that would need to be locked again, and could introduce changes in version while re-evaluating if something changed in between.

> So far as I can tell, if you use the version to alias wont that mean that only one stable package, for a given package name, can exist in artifactory at a time? The layout in artifactory is user -> package name -> version -> channel so this would only allow for one stable version at a time which seems limiting. Its highly possible I am mis understanding some element of this but can you shed some insight on this?

This is not fully clear. For the same name, version and user, yes, only 1 "stable" package can be used. Why would you want to have many different packages with exactly the same name, version, user and channel? Probably I am missing something here.

You can have in the same repo something like:

- pkg/1.0.0@user/stable (regular package)
- pkg/1.0.1@user/stable
- pkg/1.0.2@user/stable
- pkg/2.0.0@user/stable
- pkg/latest@user/stable -(alias)-> pkg/2.0.0@user/stable
- pkg/main@user/stable -(alias)-> pkg/1.0.2@user/stable
- pkg/oldest@user/stable -(alias)-> pkg/1.0.0@user/stable

Does this clarify it a bit?

Daniel-Roberts-Bose:
> * pkg/latest@user/stable -(alias)-> pkg/2.0.0@user/stable
> * pkg/main@user/stable -(alias)-> pkg/1.0.2@user/stable
> * pkg/oldest@user/stable -(alias)-> pkg/1.0.0@user/stable

Ahh I think I understand a bit, you are using an alias in this case simply as a temporary pointer that will keep moving as things change. We achieve the same behaviour with aliasing over the channel, and then using a version range to grab the "latest". 

The only potential problem I can see, is that there is no history of that "latest" pointer in your example above. Once the pointer gets moved to `pkg/latest@user/stable -(alias)-> pkg/3.0.0@user/stable` there is no longer a record that it ever pointed to `pkg/2.0.0@user/stable`. It is certainly true that in any sort of lockfile, or conaninfo file the alias wont be present and instead it'll just resolve to the actual package reference for traceability though.

I think I need to re read some of the conan documentation on code promotion in general 😆 our current flow for code promotion is that we create a reference such as `pkg/1.0.0@user/master` then once we run testing we alias the channel to `pkg/1.0.0@user/stable` to prevent us from uploading the same package multiple times, and instead just create a pointer. It sounds though from some conversation above, that uploading the same package twice isn't too much of a big deal, and is the expected practice?

ytimenkov:
We just pick channel upfront: "release" for tags and "testing" otherwise. But no special thought about it... Maybe I planned to configure Artifactory to not accept "testing" (as a pattern) in "release" repository when promoting, but haven't got there.

And we tag to create release, so release commit is built twice, but the second time it gets solid release (as in semver) version.

memsharded:
> I think I need to re read some of the conan documentation on code promotion in general 😆 our current flow for code promotion is that we create a reference such as pkg/1.0.0@user/master then once we run testing we alias the channel to pkg/1.0.0@user/stable to prevent us from uploading the same package multiple times, and instead just create a pointer. It sounds though from some conversation above, that uploading the same package twice isn't too much of a big deal, and is the expected practice?

Yes, this would be recommended, expected practice. 
The good thing is that it works efficiently in more than one dimension:
- Deduplication works at the storage level. That means that when you upload a file that already exist in another repository, not even the same repository, the storage doesn't increase. You can have the same file in multiple repositories, it will store only 1 copy.
- It is also efficient in the network transfer. If the file is already in the server, it will not only do not store an extra copy, it will avoid the upload all together.

memsharded:
This proposal was backported to Conan 1.X a few releases back, and it was a good improvement, allowed to solve some problematic corner cases, and it can be seen that it helps reducing code complexity and future maintenance. 

Most of the Tribe was not concerned about this (mostly because of not using alias at all), so there weren't many upvotes for this proposal. If you happen to use alias, please make sure to try the new syntax ``pkg/(alias)`` in latest Conan 1.X release, and submit issues to the main Conan repo. Thanks for the feedback!

PR #27, opened 2021-07-06
New package_type to complete model information

memsharded:
Conan 2.0 will introduce the optional concept ``package_type`` in recipes, to specify which kind of artifacts it contains. Possible ``package_types`` will be "shared-library", "header-library", "application", "library", etc.

- It will be an optional, non mandatory field, useful to complete information when other mechanisms fall short, for example to specify that a package contains a header-only library, options are not a good approach.
- It doesn't aim to be a complete representation or description of what is inside a package, but instead to help describing the default role of a package in a ``requires`` relationship, for example a package like ``zlib/1.2.11``, can contain executables, but its common role when ``required`` would be a library, it could have a ``package_type="library"`` (not necessary or mandatory to be specified, just an example).
- This proposal completes the model in https://github.com/conan-io/tribe/pull/26, which is the one that really describe how packages are consumed. The proposal in this PR would mainly help that other one to simplify or automate further some cases.

---


jasal82:
I'm wondering if this will solve the following use-case: Let's suppose we have a Conan recipe for an application that has a bunch of static library dependencies that are linked into the binary. Now we build and upload a binary package for that application. Would it be possible to deploy that package (e.g. via the `deploy()` method) without having to retrieve (download or build) the packages for the static library dependencies?

theodelrieu:
I've read multiple times, but I still cannot understand the problem it's trying to solve. It might be just me but I find the proposal very unclear.

memsharded:
> I'm wondering if this will solve the following use-case: Let's suppose we have a Conan recipe for an application that has a bunch of static library dependencies that are linked into the binary. Now we build and upload a binary package for that application. Would it be possible to deploy that package (e.g. via the deploy() method) without having to retrieve (download or build) the packages for the static library dependencies?

This will be possible, but not thanks to this proposal, but to the "requirement traits" proposal. Those traits will be able to model that you are not requiring transitive headers or libraries (headers=False, libs=False), when an application hides its dependencies (something like ``visible=False``). This proposal will just help to identify the package types in that relations, so for example, it might not be necessary to fully specify all traits in all ``requires()``, as long as the ``package_types`` can help to correctly complete and propagate the traits down the graph.

jasal82:
> > I'm wondering if this will solve the following use-case: Let's suppose we have a Conan recipe for an application that has a bunch of static library dependencies that are linked into the binary. Now we build and upload a binary package for that application. Would it be possible to deploy that package (e.g. via the deploy() method) without having to retrieve (download or build) the packages for the static library dependencies?
> 
> This will be possible, but not thanks to this proposal, but to the "requirement traits" proposal. Those traits will be able to model that you are not requiring transitive headers or libraries (headers=False, libs=False), when an application hides its dependencies (something like `visible=False`). This proposal will just help to identify the package types in that relations, so for example, it might not be necessary to fully specify all traits in all `requires()`, as long as the `package_types` can help to correctly complete and propagate the traits down the graph.

But where would I set the `visible=False`? The application package is not required in another conanfile, it is just being deployed via 'conan install <ref>` in my use-case.

memsharded:
> I've read multiple times, but I still cannot understand the problem it's trying to solve. It might be just me but I find the proposal very unclear.

As today, we have no model at all to represent that a package is containing a "header-only" library, for example. This brings many problems:

- There are hooks in conan center that checks for ``self.header_only()`` call, or inspect the topics, to see if a library is a header-only.
- There is a feature "full_transitive_package_id" in https://docs.conan.io/en/latest/creating_packages/define_abi_compatibility.html?highlight=transitive#enabling-full-transitivity-in-package-id-modes that is basically workarounding issues with header-only libraries
- Conan 2.0 will stop propagating libraries and headers without limit down the graph. You could fully specify in a header-only package about your dependencies that ``self.requires(....., transitive_headers=True, transitive_libs=True)``, for each one of your dependencies ( see the "requirements traits" proposal), but knowing that the package is a "header-only" package, can directly provide this information, resulting in less noisy and easier recipes.

Same happens with packages that are only a shared or a static library, which happen from time to time. An ``option = {"shared": [True]}`` works badly, is not a good solution, so there is no mechanism to know which kind of library it is. Knowing that something is shared or static is important to determine linkage requirements, for example, a shared library linking a static library does not need to propagate down the linkage requirements, but if Conan doesn't know which type of library it is dealing with, this would be impossible.  Again, this might be possible by fully specifying the ``requires traits`` for all parties involved, but as the proposal says, defining the "package_type" straight ahead can save a lot of duplication and avoid errors.

KerstinKeller:
Hi Conan Team,

thanks for this proposal. 
I like the idea that this allows a more fine-grained specification of package-id computations.
At the moment we want to move from the default `semver_mode` to `recipe_revision_mode`, but relaxing that for shared libraries seems like a reasonable idea.

However I do have some headaches with the proposal in general:

One problem I see here is how these packages exclusively categorize the packages.
This might work for `python-require` and `assets` case, but with the library/application type it's more subtle.
Only a part of the packages that we handle, can be clearly classified.
Examples would be projects that are a set of libraries (e.g. the `protobuf` package, which has a couple of libraries `libprotobuf`, `libprotoc` and a tool `protoc`. - So what would be the approriate category here? )
I would argue that these are not only edge cases, but at least one third of our packages consist of more than one library that can be clearly classified (e.g. Protobuf, Qt, Boost, OpenCV, ...)
To have an accurate model in these cases, one would need a component based classification.
I feel, that this issue might not completely work out without a revisited component modeling (also think CMake `add_subdirectory`) of some kind.

However my main headache here is that Conan needs to specify more and more information, which I consider to be build system information. From my point of view, it would be preferable, if the build system, could, in some way, communicate to Conan what it builds and packages. Thus it could somehow output what it has build (or Conan could evaluate the actual package content, and decide then and there what type of a package it is).
In short, information duplication (and discrepancies) should be avoided at all costs.
What happens if a package is declared as `header-library`, but then there are `*.dll` / `*.so` files in the `package` folder.
Will Conan issue a warning in that case? An error? 
It easy to wrongly categorize a package and the implications could be harmful and the root cause hard to locate.

memsharded:
> But where would I set the visible=False? The application package is not required in another conanfile, it is just being deployed via conan install ref together with the deploy() method implementation in my use-case.

Once the dependency graph has been computed (remember that in other proposal it was approved that the dependency graph, talking about recipes, will always be complete), the requirement traits can be used to skip unnecessary binaries and do not fetch them. Specifically, when installing the application, if it doesn't need to build from source, Conan could skip every dependency binary that doesn't have the ``run=True`` trait (that indicates it is necessary at runtime, because it could be some shared library or other file necessary for running).

KerstinKeller:
> Knowing that something is shared or static is important to determine linkage requirements, for example, a shared library linking a static library does not need to propagate down the linkage requirements, but if Conan doesn't know which type of library it is dealing with, this would be impossible

But why not find a way to let the build system figure that out, and take that info from the build system. I would argue that it's really not necessarily Conan's job to figure out how something was linked in it's function as a package manager. In an ideal world, wouldn't the following approach be more straigthforward:

1. Conan -> retrieves dependencies and forwards locations to build system 
2. Build system -> locates libraries in location that conan provides (knows if they are static / shared / ...)
3. Build system -> Links libraries to library / application
4. Build system -> communicates to conan, how things were linked, and what is the output (application, shared/static lib, ...)
5. Conan -> Gets information and makes it part of the package, so package type can be propagated to consumers

design/027-package_types.md
```diff
@@ -0,0 +1,99 @@
...
+- "application": Executable artifacts, .exe, binaries to run. It will not contain headers, static libraries, or shared libraries to be linked (might contain shared libraries used by the executable, but exclusively for the executable). Not necessary to define it, but it might be informative for recipe readers to realize that such "conanfile.py" is creating an application and not a library.
+- "python-require": This is a pure python recipe, it actually doesn’t contain any package binary or artifact at all.
+- "assets": This package contains some assets, like images, sound files, videos, fonts, configuration files, etc. They are not C/C++ headers, not libraries to link or applications to run.
+- "unknown": If the type is not defined explicitly, or cannot be deduced from options
+
+This is just the initial proposal, please report on any other potential package types that you are already managing (better not to speculate, but build on top of actual use cases) and would benefit from an explicit model.
```
DoDoENT, line 67: We currently have a package that contains only CMake helper scripts. In Conan 1.X it's advertised as `header_only`, however, in this model, it would be either `assets` or `unknown`. I'm not sure that `assets` is the best name for such package class - maybe something like `scripts` or `build-scripts`?

I'm open to suggestions...

DoDoENT:
@KerstinKeller , AFAIK this proposal is just to let the downstream know how to default-initialize traits if they are not specified. This is why this attribute will be optional. You can still have packages that pack several different things together (like `libprotobuf` and `protoc`) and you simply wouldn't specify any category in such packages. However, in the downstream, you will then need to specify in the `requires` clause what do you actually need from the package.

memsharded:
> Examples would be projects that are a set of libraries (e.g. the protobuf package, which has a couple of libraries libprotobuf, libprotoc and a tool protoc. - So what would be the approriate category here? )
I would argue that these are not only edge cases, but at least one third of our packages consist of more than one library that can be clearly classified (e.g. Protobuf, Qt, Boost, OpenCV, ...)
To have an accurate model in these cases, one would need a component based classification.

Please have a look to some of the comments in the doc, like:

> The ``package_type`` does not intend to be an exact, complete and accurate representation of every artifact that is contained in the package, in exactly the same way the current options ``shared=[True, False]``, existing in multitude of Conan packages apply only to the library/libraries contained in that package, but not to the executables that are also contained within that same package. For a packaged library that is always shared (and thus the option model does not work) a class attribute ``package_type = "shared-library"`` will be correct, even if the package contains other things like executables.

In the same way the ``shared`` option can provide good value and be useful to represent all those packages (most likely all those packages Protobuf, Qt, Boost, OpenCV already define this option), it is not necessary to have an accurate model with component classification. Actually, trying to go that way would add too much of complexity, we don't want to go there. Mainly because it is not necessary, as pointed out in the proposal, the aim is not having an accurate representation of the package, that is not necessary at all, but to just complement the "requirement traits" proposal, to help and simplify a few things.

> From my point of view, it would be preferable, if the build system, could, in some way, communicate to Conan what it builds and packages. Thus it could somehow output what it has build (or Conan could evaluate the actual package content, and decide then and there what type of a package it is).

We just merged https://github.com/conan-io/conan/pull/9005, that uses the CMakeFileAPI to try to extract information from the build. However, the information provided is very limited, not enough to provide a full package specification. And the problem is that other build systems should also provide this, but this is quite out of the scope of Conan, is really a challenging and humongous effort to try to deduce automatically from build systems what was produced. And in any case, it is not necessary to provide a detailed representation of a package for the purpose of the new graph model (requirements traits)

> What happens if a package is declared as header-library, but then there are *.dll / *.so files in the package folder.
Will Conan issue a warning in that case? An error?

What happens as today if you declare a package with a ``shared=True`` option and you package a static lib, or an executable, or only headers? What happens if you define -DBUILD_SHARED_LIBS=ON to cmake, but internally it hardcodes ``add_library(mylib SHARED)``? Or what if a ``package_id()`` method calls ``self.info.header_only()`` when the package contains not only headers? There are infinite possibilities to do these things, not only in Conan, but in every tool. In my opinion, the proposal is not adding new possible error checks like this one, but it can unveil hidden bugs, for example if now a package is incorrectly packaging some .so library and declaring ``self.info.header_only()``, the package_id will be unique, keeping the latest compilation, that could be or not compatible with others, and that will fail at runtime because of some ABI mismatch. In my experience, making things more explicit, like a ``package_type = "header-library"``, that could define the "requirement trait" ``libs=False`` will fail faster, producing an error at compilation time instead of runtime, because the library link requirement was not propagated, something that is impossible with the current model.

design/027-package_types.md
memsharded: Good case.

Is there something in the package that would condition how it is required? Would a ``build_requires`` be able to consume this package without issues?:
- header=False: No headers included
- libs=False: No libs to link
- build=True: Necessary at build time, in build context
- run=True: No need to put anything in the path, not an executable, not a shared lib. But it certainly need to be there at build time, or build will fail
- visible=False: Downstream consumers of the package are not aware how this package was built or even its build system

So maybe it is not necessary to explicitly classify it, as long as it can be consumed successfully without issues with the new ``requires`` model. If it needs to be typed, I'd probably prefer ``build-scripts``, to clearly indicate the intent.

memsharded:
> @KerstinKeller , AFAIK this proposal is just to let the downstream know how to default-initialize traits if they are not specified. This is why this attribute will be optional. You can still have packages that pack several different things together (like libprotobuf and protoc) and you simply wouldn't specify any category in such packages. However, in the downstream, you will then need to specify in the requires clause what do you actually need from the package.

Excellent point, yes, this is largely the intent. It would be very ugly to have to specify in every conan recipe:

```python
def requirements(self):
     self.requires("zlib/1.2.11", headers=True, libs=True, build=False, run=self.options["zlib"].shared, ....)
     self.requires("poco/1.9.4", headers=True, libs=True, build=False, run=self.options["poco"].shared, ...)
```

Besides, that model is not fully correct, because it is difficult to know the actual value of ``self.options["dep"].shared`` while you are expanding the graph, and such dependency has not been evaluated.

So the ``package_type``, either obtained from the current ``option.shared`` for most packages, or simply completed by the ``package_type`` for others cases where options are not enough, can greatly help to simplify recipes and let a regular ``self.requires("zlib/1.2.11")`` be ok in most of the cases, because ``zlib`` package is treated by default as a library, even if there is one executable inside that package too.

memsharded:
> But why not find a way to let the build system figure that out, and take that info from the build system. I would argue that it's really not necessarily Conan's job to figure out how something was linked in it's function as a package manager. In an ideal world, wouldn't the following approach be more straigthforward:

**In an ideal world**, that is the key point 😄 

The thing is that not only build system provides this information easily, some do not provide it at all, and it might be actually quite challenging to get it. We are far from an ideal world in the C and C++ building landscape.

It is my plan to work on a proposal for ISO C++ committee, and the ``CppPackage`` model in the conan-io/conan#9005 is a step towards it. When build systems agree on some interface that would allow to easily dump this information, as well to consume it, we will be in that much better world. But until then, this is almost an intractable problem from our side, it is simply not worth those many resources that would be necessary to develop something that would still be fragile and not cover all cases. For something that being pragmatic is adding a ``package_type = "something"`` in recipes, and that is not necessary always, just some cases.

a4z:
In what category would the iOS CMake toolchain file land , https://github.com/conan-io/conan-center-index/tree/master/recipes/ios-cmake , which is just packaging a cmake toolchain file. Unknown ? That does not look nice.
Ans what would the android-ndk be? It has something of everything in it's package.
On the other hand, how to avoid an endless growing list of package types, but keep it open for required categories that have more than 1 use case.

memsharded:
> In what category would the iOS CMake toolchain file land , https://github.com/conan-io/conan-center-index/tree/master/recipes/ios-cmake , which is just packaging a cmake toolchain file. Unknown ? That does not look nice.
Ans what would the android-ndk be? It has something of everything in it's package.

In theory it is not necessary to specify this one, as the ``build_requires`` requirement model will work fine with this kind of dependency (this will be the same issue commented by @DoDoENT above, for their cmake scripts only package). I would probably leave it without type initially, and if using and consuming it needs to define it, we could add it. Recall the main purpose of this proposal is to supply some additional information to the requires model when necessary, but not strictly required to always classify all packages under a given ``package_type``. If we realize it is necessary, I think the suggestion above of ``build-scripts`` sounds explicit and nice. 

Same would apply for the android-ndk, in theory, as a ``build_require`` it will be perfectly fine. If not, maybe creating a ``ndk`` package type that could be use for compiler-like things (executable + system/sysroot libraries + headers) could be good.

design/027-package_types.md
```diff
@@ -0,0 +1,99 @@
...
+```python
+class PackageType(Enum):
+    LIBRARY = "library"
+    HEADER = "header-library"
+    APP = "application"
+    PYTHON = "python-require"
+    UNKNOWN = "unknown"
+```
+
+To allow some type and value safety, but in general recipes should use the string format, for simplicity, while internal code might use the enum more frequently.
```
mathbunnyru, line 85: I don't believe that in general recipes should use the string format.

This is not type-safe or error-safe.
For example, one might write `shared_library` instead of  `shared-library`.

If you don't believe this happens a lot, take a look at line 19 of this document where you made this error :)

Thus, I think we have to reword a little bit this proposal.

design/027-package_types.md
```diff
@@ -0,0 +1,99 @@
...
+
+The fact that a package can define its package type as ``shared_library``, for example, does not exclude the possibility to use its packaged executables. The main mechanism to define the relationship between packages are the "Requirement traits" described in the proposal https://github.com/conan-io/tribe/pull/26. The ``package_type`` can just help that requirements model providing some information about the default role of a package in a typical ``requires`` relationship, for example, when another package is requiring ``zlib/1.2.11`` Conan is already assuming that this is a library to be consumed and linked. Build-requirements, deploys, installs, virtualenv can still be used to run executables from packages, irrespective of their ``package_type``, by defining the correct "requirement traits".
```
mathbunnyru, line 19:
```suggestion
The fact that a package can define its package type as ``shared-library``, for example, does not exclude the possibility to use its packaged executables. The main mechanism to define the relationship between packages is the "Requirement traits" described in the proposal https://github.com/conan-io/tribe/pull/26. The ``package_type`` can just help that requirements model providing some information about the default role of a package in a typical ``requires`` relationship, for example, when another package is requiring ``zlib/1.2.11`` Conan is already assuming that this is a library to be consumed and linked. Build-requirements, deploys, installs, virtualenv can still be used to run executables from packages, irrespective of their ``package_type``, by defining the correct "requirement traits".
```

mathbunnyru (changes requested):

design/027-package_types.md
memsharded: Oh, yes, it is error safe. Recall that the ``conanfile.py`` it is not pure python, Conan takes control over it.

One of the first thing it will do to is to evaluate the string to a ``PackageType`` enum, and it will automatically error if not a valid one (I have already implemented this in a proof of concept). 
I want to avoid to be forced to use:

```python
from conan.model import PackageType
from conans import ConanFile

class Pkg(ConanFile):
     package_type = PackageType.HEADER_LIBRARY
```
It requires even more knowledge about the imports and the enum.

I am confident that we can easily have the simplicity of the bare string together with the type and error safety.

mathbunnyru: Ok, thanks for the additional info 👍

DoDoENT:
Based on this idea, do you think it would also be good to define an optional, non-mandatory field `version_schema` that could be used to default initialize the versioning scheme that will be used by downstream packages when consuming this package?

Currently, you need to explicitly state the version schema in the `package_id` method for all your dependencies, which is OK if you want a specific package to influence your package's package_id in a non-default manner. However, different packages use different versioning schemas and conan always defaults to `direct_semver`. It's possible to specify [different version scheme](https://docs.conan.io/en/latest/creating_packages/define_abi_compatibility.html#versioning-schema), but this needs to be done in a downstream package, which then needs to be aware of the versioning model of your package. This can be cumbersome if you have lots of packages that may even change their versioning schemas between releases.

The proposed field would work similarly to the proposed `package_type`, except it will default-initialize the package_id method's rules in downstream packages. The downstream packages could still use their own rules in their `package_id` method, just like now - only the default wouldn't always be `direct_semver` (or whatever is set as global config), but would be set to whatever schema package advertises.

For example, let's say we have package `A` that does not use semantic versioning but makes breaking changes on every minor release (e.g. `boost`). So, if package `B` depends on `A`, currently it needs to be aware of the `A's` versioning schema and have something like this in its `package_id` method, to ensure recompilation in case downstream overrides the version of `A`:

```python
class PkgB(ConanFile):
    name = "pkg-b"
    requires = "pkg-a/1.0@user/stable"

    def package_id(self):
        myotherlib = self.info.requires["pkg-a"]
        myotherlib.version = myotherlib.full_version.minor()
```

However, if `A's` maintainer decides to change `A's` versioning schema to semantic versioning from `v2.0` onward, the users of `A` needs to be aware of that and compensate for that in their `package_id` method.

However, if package `A` could advertise its versioning scheme, then `B` could simply require it without any knowledge of `A's` maintainer decisions:

```python
class PkgA(ConanFile):
    name = "pkg-a"
    versioning_schema = 'minor'  # or something like this, could be also a function that needs to return the `info` object
```

Then the `B` could simply `require` `A` without needing to tweak `package_id`:

```python
class PkgB(ConanFile):
    name = "pkg-b"
    requires = "pkg-a/1.0@user/stable"
```

@memsharded, what do you think about it? Is this worth considering as a separate proposal?

memsharded:
@DoDoENT 

yes, I totally agree. The versioning schema of the library is something that we want to improve for Conan 2.0. There are 2 different points to this proposal though:

- The versioning schema, that affects how versions and version ranges are resolved. This is something that I would like to see improved to be able to depend on latest ``openssl`` despite the ugly letter that breaks semver.
- The default ABI changes that some versioning scheme would imply by default to consumers.

Both are worth addressing, probably in that order, first define different types of versioning, and later, based on those version types, define ABI influence. Lets do it later in other proposals.

sztomi:
I think this is an excellent proposal and will make our lives a lot easier. 

> visible (default True): This require will be propagated downstream, even if it doesn’t propagate headers, libs or run traits. Requirements that propagate downstream can cause version conflicts. This is by default True, because in most cases, having 2 different versions of the same library in the same dependency graph is at least complicated, if not directly violating ODR or causing linking errors.

Is  `visible=False` intended to replace private requires? Our use-case is the following: we have auxiliary executables that are distributed as conan packages. These are bundled in the final build of our product (so we pretty much just copy things from the package dir to the build dir). For a long time, we used to use build_requires to hide the dependencies of these auxiliary packages (so as to not require _their_ dependencies transitively in the final consuming conanfile). For example, if we required a statically linked python executable package, we did not want to add bzip2 to the dependency graph. But this is a misuse of  build_requires and it has its own problems, particularly with package revisions.

ohanar:
I think addressing the @DoDoENT's CMake script package, or other sorts of non-standard packages should be handled here in an extensible manner.

Assuming I'm understanding this PR correctly, this basically exists to provide default values for consumers for the requirements traits in #26. Given that, I think the simplest option for extension would be to be by allowing package_type to be a dictionary of defaults. Then the standard package types are just shortcuts for common default dictionaries. So for example, you would have something along the lines of
```python
package_type = "header-library"
# is equivalent to
package_type = {
    "headers": True,
    "libs": False,
    "run": False,
    "transitive_headers": True,
    "transitive_libs": False,
    ...
}
```
This also makes things like testing frameworks (e.g. catch2) specify that they should be test includes by default. Given that the package type can be specified during the configure method, that would allow the dictionary to have fields set based off of options (this would be how one must implement the "library" package type using a dictionary).

design/027-package_types.md
ytimenkov: We have agreed to use python 3.6 with type annotations. It is possible to specify a literal union type (same for `mode` in `open()`:

```python
package_type: Optional[Literal["library", "asset", ...]]
```

Then most of type checkers will catch mistakes. It is still a good idea to pass recipe through a linter and type checker in a pre-export hook.

ytimenkov:
> However my main headache here is that Conan needs to specify more and more information, which I consider to be build system information. From my point of view, it would be preferable, if the build system, could, in some way, communicate to Conan what it builds and packages.

This is a chicken-and-egg problem: Conan needs to analyze information before passing it further down to the build system. Simple example: depending on `shared` option you may get different set of requirements. Or depending on options you may get a different set of targets: i.e. information flows other way around: instead of build system telling Conan what it has built, it's Conan tells build system what to build (usual mapping from `WITH_*` or `USE_*` or `ENABLE_*` defines / autoconf flags into Conan options).

Well, probably the latter is a lesser of problem for consuming - you may have built package when populating `self.cpp_info`, but still there is a huge overlap of information.

But I agree that it would be nice to reduce the overlap and duplication.

memsharded:
@ohanar 

The problem with your proposal is that it would make its declaration more complicated, and not that evident. For example, one package application like cmake will need to declare:

```
package_type = {
    "headers": False,
    "libs": False,
    "run": True,
    ...
}
```
And maybe others, because the defaults are not the expected ones. I think this is less evident that saying ``package_type = "application"``.

Furthermore, there are some properties that cannot be declared by the creator of the package, for example ``transitive_headers`` and ``transitive_libs``. This is something that belongs exclusively to the relation between the consumer and the package, and only the consumer can declare these, because it is the only who knows how it is propagating downstream the dependency headers and libs. Something similar happens for the ``test`` trait, gtest might be used by other testing framework **without** the ``test`` trait, because it is extending or using it, not testing itself with it, while regular libraries like Boost might be used as a test requires. too. And other traits like ``visible``, ``force``, ``override`` directly make no sense in this package-type, so I am not sure the apparent similarity between this proposal and the ``requirement traits`` will bring more confusion than the value it brings with a more complete representation.

ohanar:
> The problem with your proposal is that it would make its declaration more complicated, and not that evident. For example, one package application like cmake will need to declare:
> 
> ```
> package_type = {
>     "headers": False,
>     "libs": False,
>     "run": True,
>     ...
> }
> ```
> 
> And maybe others traits too, because the defaults trait values are not the expected ones. I think this is less evident than saying `package_type = "application"`.

I wasn't proposing "instead of" but rather "in addition to". The vast majority of the time, a user would use one of the package types described in the proposal, this is mainly to address strange edge cases, where something more is needed.
 
> Furthermore, there are some properties that cannot be declared by the creator of the package, for example `transitive_headers` and `transitive_libs`.

The package_type is just specifying the default and can always be overridden by a consumer. I can think of at least one case of a package we have internally that it would make sense for `headers != transitive_headers` for the default for a consumer. Again, it is a weird package, so it doesn't fit under the clean categories given in the proposal, hence the need for more flexibility than what the proposal currently lays out.

> Something similar happens for the `test` trait, gtest might be used by other testing framework **without** the `test` trait, because it is extending or using it, not testing itself with it,

Again, defaults can always be overwritten by consumers. How frequently is gtest as a testing tool vs by another testing framework as a library/application? If gtest were almost always used as a library/application by other testing frameworks, `test = False` should be the default, but that is the exception, and not the general rule, so `test = True` should be the default, and those frameworks who need it as something different would overwrite that.

> And other traits like `visible`, `force`, `override` directly make no sense in this package-type, so I am not sure the apparent similarity between this proposal and the `requirement traits` will bring more confusion than the value it brings with a more complete representation.

I think this is more of an issue of lumping too many things under the requirement traits umbrella without clear documentation categorizing them. There are those that mostly affect the dependency graph (visible, force, override, build), vs those that provide metadata for recipes+generators (libs, headers, run, direct, test, transitive_headers, transitive_libs). I think that with an appropriate categorization, a dictionary package_type could specify the default values for consumers' "metadata requirement traits", with the convenience of using "header-library", "library", "application", etc. for common scenarios.

design/027-package_types.md
lasote: @memsharded as we agreed, add the `build-scripts` to the proposal. Introduced here: https://github.com/conan-io/conan/pull/10076

lasote (approved):

memsharded:
Time to merge this PR. The overall feedback seems positive, suggestions have been incorporated already to the code implementing this PR, and initial feedback of the 2.0-alpha releases is positive too. Please name possible issues related to this new package_type in the main Conan repo as ``[develop2]``.

Thanks all for the many useful discussions and ideas!

PR #28, opened 2021-09-16
Middleware to customize all recipes

gmeeker:
A proposal to provide support for advanced customizations like multiple architecture support.

### DIFF
--- a/design/###-middleware.md
+++ b/design/###-middleware.md
@@ -1,0 +1,153 @@
+# Proposal: Middleware to customize all recipes
+
+| **Status**        | **Proposed** |
+|:------------------|:---------------------------------------------|
+| **RFC #**         | ####                                         |
+| **Submitted**     | 2021-09-15                                   |
+
+---
+
+## Summary
+Allow users to inject a layer or layers of code between Conan and all recipes.
+For example, this would allow users to customize the build() and package() methods,
+even for the recipes in CCI.  This provides support for advanced features
+like multiple architectures without requiring any commitment to a particular approach.
+
+
+## Motivation
+Conan currently does not support multiple architecture (multiarch) binaries.
+iOS and macOS call these universal binaries (and the tool is named *lipo*).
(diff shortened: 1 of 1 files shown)


memsharded (commented):
Hi @gmeeker 

Thanks very much for your detailed proposal.

However, it seems this is too much at this stage an incubating idea, rather than a 2.0 proposal. (Please read the notes in https://github.com/conan-io/tribe#important). If you have a look to what has already been discussed here is about stabilization, deprecation and removal of unused or legacy features, etc:
- What are the minimum versions to support.
- Using revisions always (revision have been implemented as opt-in for years, and used in production by many users)
- Very well known limitation of the graph model
- The well known problem of mixing cases in package references (ConanCenter has been running on lowercase for years too)
- Deprecating unused features, like the manifest checker.

Most of the "new features" that are going to be implemented for 2.0 are already incubated in 1.X, for example all the new build integrations, the layout feature to help with the "editable" issues, or the new environment management: https://docs.conan.io/en/latest/reference/conanfile/tools.html.

It seems that this middleware concept is too much of a new, big change, and should be incubated first in 1.X or in 2.X, if it makes sense, introduced as an opt-in, gain experience, etc., in order to gain feedback from other users as well. 

I suggest moving this to the Conan client main repo, try to break it into smaller pieces and try to progress there.

However, I don't think we have the resources to implement these new features on a reasonable time for 2.0, we are already struggling with implementing very fundamental changes, like making the Conan cache multi-revision (challenge bigger than expected), or improving very core things as the dependency resolution models and algorithms.


Also, if you are interested in the Tribe work, it would be good to explicitly become a member, you can write us an email, we might be adding new members to it.

gmeeker:
Sorry, I'm not sure how I missed that.  Yes, I'll move over the the main repo.  I'm still pretty new to Conan so I don't presume to know how others would use this middleware concept.

memsharded:
No prob @gmeeker, thanks! 😃

PR #31, opened 2022-01-17
Remove "compiler.base" settings mechanism

memsharded:
The mechanism of defining that a compiler such as “intel” in the settings.yml definition will kind of extend other existing compilers (such as “intel” having as a base either “gcc” in Linux or “msvc” in Windows) will be removed. Instead, a full definition of the compiler will be necessary, as the new IntelOneApi (via ``intel-cc`` setting has already been proposed in Conan 1.X).

The potential expected benefits of the "compiler.base" mechanism are not worth it compared with the complexity they bring. Some users, the ones that promoted this feature have already requested its removal. It also seems to have no extended usage beyond the previous ``intel`` and ``mcst-lcc`` compilers.

---

### DIFF
--- a/design/031-remove_compiler_base.md
+++ b/design/031-remove_compiler_base.md
@@ -1,0 +1,41 @@
+# Proposal: Remove "compiler.base" settings mechanism
+
+
+| **Status**        |                                                   |
+|:------------------|:--------------------------------------------------|
+| **RFC #**         | [031](https://github.com/conan-io/tribe/pull/31)  |
+| **Submitted**     | 2022-01-17                                        |
+| **Tribe votes**   |                                                   |
+
+
+## Summary
+
+The mechanism of defining that a compiler such as “intel” in the settings.yml definition will kind of extend other existing compilers (such as “intel” having as a base either “gcc” in Linux or “msvc” in Windows) will be removed. Instead, a full definition of the compiler will be necessary, as the new IntelOneApi (via ``intel-cc`` setting has already been proposed in Conan 1.X).
+
+
+## Motivation
+
+When this feature was introduced, following feedback of users mainly using the Intel compiler, it seemed useful. Such users were reporting binary compatibility between packages built with Intel and packages built with the underlying gcc or Visual compiler, so, apparently, having a built-in definition of such base compatibility made sense.
+
+But, since then, many of these users have now reported that such a feature is not adding value at all, and in fact in some cases it might be more confusing and problematic than the problem it tries to solve. Users are now directly defining specific settings per-package to explicitly define which configuration they want, and following their feedback, the new ``intel-cc`` compiler setting for IntelOneApi was defined, without such mechanism.
+
+Removing the feature will simplify the implementation of generators and toolchains, and users wanting to implement some kind of fallback compatibility between different compilers will be able to use other mechanisms.
+
+
+## Detailed Design
+
+Note that this proposal doesn’t remove any ``compatible_packages`` mechanism. Actually, we are considering how to better represent such compatibility, with other better UX than coding it directly in the recipes ``package_id()`` method. This proposal only removes the ``settings.yml`` definition of the ``intel`` and ``mcst-lcc`` compilers, and replace them with a fully self-contained one, adding all the details they need to represent their binaries, without resorting to the definition of another compiler.
+
+
+## Implementation details
+
+The ``settings.yml`` file will be updated with the removal of the “base” subsettings, and the generators and code will remove all the checks of ``self.settings.compiler.base``.
+
+
+## Migration plan
+
+The Intel compiler already got a new ``intel-cc`` new compiler in 1.X that is fully independent. While it only supports at the moment IntelOneApi, it can be extended for previous versions if necessary.
+
+The same process needs to be done for the ``mcst-lcc`` compiler in 1.X, defining a new independent one.
+
+The resulting ``package_ids`` are different, so for the new settings there shouldn’t be any conflict with binaries.
--- a/design/031-remove_compiler_base.md
+++ b/design/031-remove_compiler_base.md
@@ -10,7 +10,11 @@
 
 ## Summary
 
-The mechanism of defining that a compiler such as “intel” in the settings.yml definition will kind of extend other existing compilers (such as “intel” having as a base either “gcc” in Linux or “msvc” in Windows) will be removed. Instead, a full definition of the compiler will be necessary, as the new IntelOneApi (via ``intel-cc`` setting has already been proposed in Conan 1.X).
+The ``compiler.base`` subsettings mechanism will be removed. Some compilers such has ``intel`` define in *settings.yml* a subsetting called ``base``, that can point to another existing compiler like ``gcc`` or ``msvc`` to which
+it assumes binary compatibility. It doesn't replicate the whole ``gcc`` or ``msvc`` subsettings, but instead uses
+a yaml syntax (e.g: ``<<: *visual_studio``). It also assumed full binary compatibility between the "intel" compiler and the "base" compiler (being it Visual in Windows and gcc in Linux).
+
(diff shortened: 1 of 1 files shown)


design/031-remove_compiler_base.md
```diff
@@ -0,0 +1,41 @@
...
+| **Status**        |                                                   |
+|:------------------|:--------------------------------------------------|
+| **RFC #**         | [031](https://github.com/conan-io/tribe/pull/31)  |
+| **Submitted**     | 2022-01-17                                        |
+| **Tribe votes**   |                                                   |
+
+
+## Summary
+
+The mechanism of defining that a compiler such as “intel” in the settings.yml definition will kind of extend other existing compilers (such as “intel” having as a base either “gcc” in Linux or “msvc” in Windows) will be removed. Instead, a full definition of the compiler will be necessary, as the new IntelOneApi (via ``intel-cc`` setting has already been proposed in Conan 1.X).
```
sztomi, line 13: I can't parse this sentence, it reads like it was half-rewritten and something remained there?

memsharded: Probably the parenthesis is wrong, would 
```
 Instead, a full definition of the compiler will be necessary, as the new IntelOneApi (via ``intel-cc`` setting) has already proposed in Conan 1.X.
```
be better?

design/031-remove_compiler_base.md
```diff
@@ -0,0 +1,41 @@
...
+But, since then, many of these users have now reported that such a feature is not adding value at all, and in fact in some cases it might be more confusing and problematic than the problem it tries to solve. Users are now directly defining specific settings per-package to explicitly define which configuration they want, and following their feedback, the new ``intel-cc`` compiler setting for IntelOneApi was defined, without such mechanism.
+
+Removing the feature will simplify the implementation of generators and toolchains, and users wanting to implement some kind of fallback compatibility between different compilers will be able to use other mechanisms.
+
+
+## Detailed Design
+
+Note that this proposal doesn’t remove any ``compatible_packages`` mechanism. Actually, we are considering how to better represent such compatibility, with other better UX than coding it directly in the recipes ``package_id()`` method. This proposal only removes the ``settings.yml`` definition of the ``intel`` and ``mcst-lcc`` compilers, and replace them with a fully self-contained one, adding all the details they need to represent their binaries, without resorting to the definition of another compiler.
```
mrjoel, line 27: > Note that this proposal doesn’t remove any ``compatible_packages`` mechanism.

Thanks for this clarification, that was my main concern in reading this. In particular wanting to make sure that this doesn't preclude package compatibility between e.g. Clang on Win32, `clang-cl`, and MSVC. Sounds like that will not be precluded but just the common compiler bits will need to be intentionally duplicated where appropriate, for which being more explicit seems a net win to me.

memsharded: In fact, better mechanisms for defining global binary compatibility are being investigated, check for example: https://github.com/conan-io/conan/pull/10241. Indeed this proposal is only talking about ``compiler.base`` implicit definition of compiler compatibility.

However, it is not fully clear what you comment for ``clang-cl`` there was never a built-in binary compatibility definition between clang-cl and msvc, do you mean you implemented yours?

design/031-remove_compiler_base.md
sztomi: Oh sorry, I meant the first sentence:

```
The mechanism of defining that a compiler such as “intel” in the settings.yml definition will kind of extend other existing compilers (such as “intel” having as a base either “gcc” in Linux or “msvc” in Windows) will be removed.
```

memsharded: Got it, will rewrite it right now to clarify

memsharded: Done, please re-check https://github.com/conan-io/tribe/pull/31/commits/08a7c01e954f2924cd9f5797ee1f77c63377f407 proposal to see if more clear now.

design/031-remove_compiler_base.md
mrjoel: > Indeed this proposal is only talking about compiler.base implicit definition of compiler compatibility.

👍 💯 

> clang-cl there was never a built-in binary compatibility definition between clang-cl and msvc, do you mean you implemented yours?

Not implemented or even explored yet much, but's it's cropped up as a roadmap item we'll need to address internally in the future. Thus my just wanting to make sure it wasn't precluded, not needed to be currently supported.

design/031-remove_compiler_base.md
sztomi: Yeah that's much better, thanks!

design/031-remove_compiler_base.md
puetzk: Another use case where I would have liked to use this sort of thing is the case of compiler wrappers that have significant binary implications, e.g. [winegcc](https://www.winehq.org/docs/winegcc). In some ways, viewing these as a separate compiler is appropriate (due to the unusual options and linker behavior, they do produce binaries which are not in general usable by the base toolchain). But, at least in the case of winegcc it *is* possible and relatively normal to link to requirements that were built by the underlying gcc toolchain.

However, compiler.base in its current form certainly was not enough for this, and removing it doesn't hurt our workarounds any (I just made a separate `winelib` setting, which the wine-aware recipe require be set to `True`, and other recipe do not include in their `settings` and therefore ignore completely).

The related [`base_compatible` and `parent_compatible`](https://docs.conan.io/en/latest/creating_packages/define_abi_compatibility.html?highlight=compiler%20base#compatible-compilers) mechanism
would have been a more fluent way to declare this sort of compatibility, but they are not very usable in conan 1.x because of other limitations.
1. compatible_packages has to be patched into each producer recipe (who has no reason to knows anything of oddities like [wine](https://winehq.org) to declare that were `parent_compatible` with something looking for a winegcc binary. So this would preclude any easy re-use of conan-center recipes.
2. This would be a mis-use of `compatible_packages` as it is not "interchangeable at all effects" - The compatibilty is a one-way thing (winegcc can safely require packages built with the base gcc, not vice-versa)
3. There's no way for the consumer to actually *prefer* that a given requirement be a package built with the base compiler. A profile can (with package-specific settings), but this gets into the usual mess that causes if there are transitive dependencies.
4. any actual use of compatible_packages where the fallback is actually used renders offline builds impossible, because the absence of a specific package_id is not cachable. So it will *always* want to contact an artifactory remote to look for the exact-match, and error out when the remote is unreachable (e.g. off the VPN).

So while describing winegcc would perhaps have benefited from something along the lines of `parent_compatible(compiler="winegcc"))` (if compatible_packages had been more flexible), I don't see this use case as much of a reason to argue for keeping `compiler.base`. If other issues with compatible_packages are later improved up, it would be easy enough to supply a suitable `winegcc_compatible` transformation in some other way, e.g. by wine's package via `python_requires`.

memsharded: Thanks for the feedback @puetzk, very useful. I think related to your point 1) you might also be interested in tracking the idea in https://github.com/conan-io/conan/pull/10241. 

I also would like to ask to which level you use per-package explicit settings in your profile, like ``specificpkg:compiler=winegcc`` for example (this can be quite powerful in profiles with jinja templates if you have some more complex logic).

puetzk: Currently, we do occasionally use per-package settings, but only a last resort because such overrides do not propagage down the transitive requirement graph (e.g. If I want a non-winegcc libxml, and it wants zlib, that needs to be a non-winegcc zlib as welll). Therefore, any case where they are used would cause not only the affected package, but every single one of its users, to have to encode all the settings overrides for of all transitive requirements, including implementation details it would otherwise have no reason to know about.

When we do have to resort to it, the structure is usually a package with a conanfile.py which supports only export-packge, and a CMakeLists.txt that uses cmake-conan to generate a conanfile.txt (using `REQUIRES`) and profile (using `SETTINGS`) for its dependencies, and then we conceal the requirements receiving overriddens ettings entirely by failing to mention them in the conanfile.py. This is generally a manageable hack, since the case for overriding from winegcc to gcc is usually using a library that will already be in the target's rootfs, but it's not really ideal. I will look into conan-io/conan#10241.

memsharded:
There has been no issues with this proposal, seems everybody is aligned with the users requests that started this proposal. This will be in the next alpha.4

PR #35, opened 2022-05-17
[Update proposal] Propose Python minimum version support updates

memsharded:
Hi Tribe!

The Tribe proposal in PR https://github.com/conan-io/tribe/pull/3 about "using a fixed Python minimum supported version" is becoming outdated at the fast pace that the Python foundation is implementing for their releases (see https://endoflife.date/python). The minimum Python 3.6 that was agreed back then has already been declared End Of Life (EOL). Keeping Conan running with unsupported Python versions suppose and increasing challenge as time passes, involving breaking dependencies (Conan dependencies no longer supported EOL versions), and security risks if those dependencies cannot be updated.

Therefore, we propose a moving minimum supported Python version, that will extend 12 months over the official Python EOL. That means that Conan 2.0 launch will still support Python 3.6, but some time after it launch, in December 2022, it might start to require Python 3.7. 

We know that this could be a challenging requirement for some orgs that can't upgrade Python versions easily, so as always we will try our best to extend as much as possible the backwards compatibility. But at some point it can really become a burden for evolution and a security liability.

---

### DIFF
--- a/design/003-codebase-python.md
+++ b/design/003-codebase-python.md
@@ -9,7 +9,8 @@
 ---
 
 ## Summary
-Declare **Python 3.6** as the minimum Python version supported.
+Declare **Python 3.6** as the minimum Python version supported for Conan 2.0 launch.
+Declare a Python version support policy for 12 months after the minimum supported version has officially become End Of Life (EOL).
 
 
 ## Motivation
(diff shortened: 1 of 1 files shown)


ytimenkov:
> We know that this could be a challenging requirement for some orgs that can't upgrade Python versions easily

Trying to have so low minimum version will be challenging for Conan as well, even if you decide to not add new language features into the codebase (like `async` or pattern matching)
1. You need to run CI and it will be challenging to find an image which has old python.
2. Even if you use own image which you don't update for CI, Conan uses a number of dependencies which (as you mentioned) could break any time. Most users won't be able to install it anyways.

Like trying to install on CentOS 6 - nothing works- old python, old pip (can't talk to pypi.org), old SSL (uses weak ciphers so servers refuse to talk) 🤦‍♂️.

I would say 12 months **after** EOL is quite generous and you are not Microsoft to patch Windows XP for EternalBlue...

